PR #13, opened 2019-08-07
Add a new Task List

danrabbit:
I think this works well enough for an MVP of adding new task lists


marbetschar:
the thrown error:

```
** (io.elementary.tasks:121011): CRITICAL **: 09:30:19.769: TaskModel.vala:74: Unable to connect to “abc”: Backend factory for source “e48bb86280db80ead228459360822830b4053b9b” and extension “Task List” cannot be found.
```

indicates, that we need to set the right values for the [`E.SOURCE_EXTENSION_TASK_LIST`](https://valadoc.org/libedataserver-1.2/E.SOURCE_EXTENSION_TASK_LIST.html) on the newly created `E.Source`.

Because the Calendar app this is done like [this](https://github.com/elementary/calendar/blob/b31b004acbfe2d0729c6e7876893062d18cf7c1a/core/Backends/LocalBackend.vala#L38-L50), I changed your code in Tasks accordingly. Now it works for adding/deleting tasks list - although only for local ones.

src/MainWindow.vala
```diff
@@ -190,6 +213,35 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+                registry = Tasks.Application.model.get_registry.end (res);
+            } catch (Error e) {
+                critical (e.message);
+                return;
+            }
+
+            try {
+                var new_local_source = new E.Source (null, null);
+                E.SourceTaskList source_task_list = (E.SourceTaskList)new_local_source.get_extension (E.SOURCE_EXTENSION_TASK_LIST);
+                source_task_list.backend_name = "local";
```
danrabbit, line 229: There must be a different name for this because I'm getting duplicate headers when adding a new list. There's `E.SOURCE_EXTENSION_LOCAL_BACKEND` but this seems to be something different and gives us a bunch of errors if we try to use it

marbetschar: @danrabbit can you elaborate on what you mean by "duplicate headers"...? I can't see any error/warning outputs using `export G_MESSAGES_DEBUG=all`.

FWIW: I just fixed the two `(io.elementary.tasks:146414): GLib-CRITICAL **: 07:34:25.964: g_utf8_collate: assertion 'str2 != NULL' failed` errors in my latest commit. Apparently we also need to set a `display_name` and a `parent` for the source.

danrabbit: Nice catch. I was thinking the sort func probably needs to be able to handle null named lists in case we pull one from elsewhere.

Are you not seeing this?: 
Screenshot from 2020-10-08 10 35 52@2x

danrabbit: Ah this must have to do with invalidating headers because it's fixed on restarting the app

src/MainWindow.vala
```diff
@@ -44,15 +46,19 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
     static construct {
         Hdy.init ();
 
+        action_accelerators[ACTION_ADD_NEW_LIST] = "<Control>N";
```
marbetschar, line 49: Do you really think we need an accelerator for creating a new list? IMHO this is quite a rare action - and I'd think if we use this shortcut, it would be more usefull for adding a new task?

marbetschar (approved):
Nice job! Feel free to merge if you disagree with my `Ctrl + N` concerns ;)

src/MainWindow.vala
marbetschar: Put some more thoughts into this, the `<Control>N` shortcut may come in handy for: https://github.com/elementary/tasks/issues/148. Therefore resolving and merging this 🥳

PR #15, opened 2019-08-07
MainWindow: Add warning before deleting

danrabbit:
Is it better to throw a warning here because this is not really undo-able or should we throw a toast and delay deleting until the toast times out?

Screenshot from 2019-08-07 16 56 29@2x


cassidyjames:
Toast and delay imho

marbetschar:
@danrabbit FYI: the new `Calendar.Store` provides a [trash API](https://github.com/elementary/tasks/blob/4ec8d47dc0c67f6bafea8851e4ca8496e2815777/src/Services/Calendar/Store.vala#L170-L191). Once merged, we can integrate the preferred UX approach suggested by @cassidyjames

danrabbit:
Superseded by https://github.com/elementary/tasks/pull/279

PR #17, opened 2019-08-08
fix #16

### LINKED ISSUE #16: Use a popover instead of a dialog to change the properties of each task list
Why is a dialog used to change the properties of a task list? Do we want to edit many task list at the same time?

I would prefer to use a Gtk.Popover for this purpose, in this way we make the editing of the task list somewhat secondary and not overlap the main purpose of completing tasks.

I will present a PR to see my idea.

alainm23:
image

### DIFF
--- a/src/ListView.vala
+++ b/src/ListView.vala
@@ -30,14 +30,15 @@ public class Tasks.ListView : Gtk.Grid {
         label_style_context.add_class (Granite.STYLE_CLASS_H1_LABEL);
         label_style_context.add_class (Granite.STYLE_CLASS_ACCENT);
 
-        var settings_button = new Gtk.ToggleButton ();
+        var list_settings_popover = new Tasks.ListSettingsPopover ();
+        
+        var settings_button = new Gtk.MenuButton ();
         settings_button.valign = Gtk.Align.CENTER;
-        settings_button.add (new Gtk.Image.from_icon_name ("view-more-horizontal-symbolic", Gtk.IconSize.MENU));
         settings_button.tooltip_text = _("Edit Name and Appearance");
+        settings_button.popover = list_settings_popover;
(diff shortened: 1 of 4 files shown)


src/ListView.vala
```diff
@@ -30,20 +30,30 @@ public class Tasks.ListView : Gtk.Grid {
         label_style_context.add_class (Granite.STYLE_CLASS_H1_LABEL);
         label_style_context.add_class (Granite.STYLE_CLASS_ACCENT);
 
-        var settings_button = new Gtk.Button.from_icon_name ("view-more-horizontal-symbolic", Gtk.IconSize.MENU);
+        var settings_button = new Gtk.ToggleButton ();
```
danrabbit: Instead of a ToggleButton, can we use a MenuButton here? Then we won't have to manually activate and deactivate the button and manually hide/show the popover

src/Widgets/ListSettingsPopover.vala
```diff
@@ -141,65 +147,65 @@ public class Tasks.ListSettingsDialog : Gtk.Dialog {
...
-            destroy ();
-        });
+        name_entry.changed.connect (save);
+    }
+
+    private void save () {
+        _source.display_name = name_entry.text;
+        try {
+            _source.write.begin (null);
+        } catch (Error e) {
```
danrabbit: I'm getting compiler output that this is an unreachable catch

src/ListView.vala
alainm23: You are right.

src/Widgets/ListSettingsPopover.vala
```diff
@@ -1,133 +1,138 @@
...
+                    break;
+            }
+        }
+
+        get {
+            return _source;
+        }
+    }
+
+    public ListSettingsPopover () {
```
danrabbit: Can we move these properties to the `construct` block since they're not constructor arguments?

alainm23: Of course.

PR #24, opened 2019-10-15
Load Tasks

marbetschar:
WIP pull request to load and display Tasks.


src/Widgets/ListRow.vala
```diff
@@ -115,4 +134,16 @@ public class Tasks.ListRow : Gtk.ListBoxRow {
...
+
+    private void on_objects_added (E.Source source, ECal.Client client, SList<unowned ICal.Component> objects) {
+        stdout.printf("on_objects_added");
+    }
+
+    private void on_objects_modified (E.Source source, ECal.Client client, SList<unowned ICal.Component> objects) {
+        stdout.printf("on_objects_modified");
+    }
+
+    private void on_objects_removed (E.Source source, ECal.Client client, SList<unowned ECal.ComponentId> uids) {
```
danrabbit:
```suggestion
    private void on_objects_removed (E.Source source, ECal.Client client, SList<unowned ECal.ComponentId?> uids) {
```

Should fix a build error :)

danrabbit:
Copy-pasting from Slack:

so now you get a list of ICal.Components on load. Try this out :wink:
```
    private void on_objects_added (E.Source source, ECal.Client client, SList<unowned ICal.Component> objects) {
        objects.foreach ((component) => {
            critical (component.get_summary ());
        });
    }
```

So I think what we need to do is find someplace to store this info. it looks like in Calendar what we're doing is we have a CalendarModel class and we're creating a hashtable like this:
```
        public HashTable<E.Source, Gee.TreeMultiMap<string, ECal.Component>> source_events { get; private set; }
```

And then when ListView has its source property updated, we can get this multimap of components and fill a listview there

To be specific, on the Gtk side what you want is https://valadoc.org/gtk+-3.0/Gtk.ListBox.html

marbetschar:
Currently implementing the `TaskListModel` and facing the following error:

```
error: The name `get_rid' does not exist in the context of `ECal.ComponentId'
```

I assume because in `meson.build` the dependency `libecal-1.2` is listed, instead of `libecal-2.0`. But if I change it to `libecal-2.0` I get:

```
error: The name `get_rid' does not exist in the context of `ECal.ComponentId'
```

**How to install `libecal-2.0`?**

danrabbit:
@tintou can you offer some guidance here?

src/Widgets/TaskRow.vala
```diff
@@ -33,6 +33,9 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         var label = new Gtk.Label (component.get_summary ());
         label.wrap = true;
         label.xalign = 0;
+        if( component.get_status () == ICal.PropertyStatus.COMPLETED ){
+            label.set_markup ("<span foreground=\"gray\">" + component.get_summary () + "</span>");
+        }
```
ryonakano:
```suggestion
        if (component.get_status () == ICal.PropertyStatus.COMPLETED){
            label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
        }
```

In order to support dark mode

marbetschar:
This draft PR became obsolete and is replaced by the following PR's:

- https://github.com/elementary/tasks/pull/43
- https://github.com/elementary/tasks/pull/38

PR #33, opened 2019-10-16
TaskRow: style past due tasks

danrabbit:
Screenshot from 2019-10-22 11 17 53@2x


marbetschar:
@danrabbit wondering if it is enough to just make the due date label (maybe just in the list color?) - without any additional icons etc.

PR #39, opened 2019-10-17
Add task description in tasklist

### LINKED ISSUE #35: Add task description in task list view
right below the task summary - in a lighter font color. Similar to what Apple does in their reminder app: Without line breaks and just the first n characters. This makes it easier to work with a lot of tasks.

marbetschar:
This fixes #35.

How to replace multiple linebreaks with just one? I expect my solution does not under all cirumstances, but didn't found a regex function for string.replace.

### DIFF
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -44,7 +44,11 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         var description = component.get_description ();
         if( description != null ) {
             description = description.replace("\r", "").replace("\n\n", "\n").strip();
-            description = string.joinv(" ", description.split("\n"));
+            string[] lines = description.split ("\n");
+             string stripped_description = lines[0].strip ();
+             for (int i = 1; i < lines.length; i++) {
+                 stripped_description += " " + lines[i].strip ();
+             }
 
             if( description.length > 0 ){
                 var description_label = new Gtk.Label (description);
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -43,15 +43,19 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
 
         var description = component.get_description ();
         if( description != null ) {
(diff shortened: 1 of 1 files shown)


src/Widgets/TaskRow.vala
```diff
@@ -30,16 +30,33 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
         grid.margin_start = grid.margin_end = 24;
         grid.column_spacing = 6;
         grid.add (check);
-        grid.add (label);
+        grid.add (summary_label);
+
+        var description = component.get_description ();
+        if( description != null ) {
+            description = description.replace("\r", "").replace("\n\n", "\n").strip();
+            description = string.joinv(" ", description.split("\n"));
```
danrabbit:
```suggestion
            string[] lines = description.split ("\n");
             string stripped_description = lines[0].strip ();
             for (int i = 1; i < lines.length; i++) {
                 stripped_description += " " + lines[i].strip ();
             }
```

src/Widgets/TaskRow.vala
```diff
@@ -30,16 +30,33 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+
+            if( description.length > 0 ){
+                var description_label = new Gtk.Label (description);
+                description_label.wrap = true;
```
danrabbit: We probably don't want this right? If we're setting `lines = 1` then this is never used

src/Widgets/TaskRow.vala
```diff
@@ -30,16 +30,33 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+                description_label.xalign = 0;
+                description_label.lines = 1;
+                description_label.ellipsize = Pango.EllipsizeMode.END;
+                description_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+
+                grid.attach_next_to (description_label, summary_label, Gtk.PositionType.BOTTOM);
```
danrabbit: Also a great solution!

PR #43, opened 2019-10-19
Use a lighter tone for the completed label summary text.

marbetschar:
To make them easier to distinguish from the ones which are incomplete in a list where items of both states are shown.

### DIFF
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -38,7 +38,7 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         label.wrap = true;
         label.xalign = 0;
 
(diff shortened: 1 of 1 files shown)


src/Widgets/TaskRow.vala
```diff
@@ -38,6 +38,10 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         label.wrap = true;
         label.xalign = 0;
 
+        if (component.get_status () == ICal.PropertyStatus.COMPLETED) {
```
danrabbit: We already check for completion at the beginning of the `construct` block :)

```suggestion
        if (completed) {
```

marbetschar: Good catch! ;)

PR #50, opened 2019-10-20
Output due date where it is set

### LINKED ISSUE #49: Show due date in list if any
## Prerequisites

## Feature
It would be beneficial to show the due date of a task in the list view. This way the user can see due tasks at a glance without going to any detail view of each task first.

If a due date is set, it should be shown after summary on a new line - between summary and description (if any). It should be formatted in the users locale.

**Additional context**
We may also consider to accentuate this output if the task is overdue with the lists color. See https://github.com/elementary/tasks/issues/47

marbetschar:
This fixes #49.

Open task:
- [x] Is there a way we can output the time in the users preferred locale, but without seconds? I think the seconds output is confusing and don't see any added value in it:

Screenshot from 2019-10-20 11 07 36@2x

### DIFF
--- a/data/TaskRow.css
+++ b/data/TaskRow.css
@@ -1,0 +1,24 @@
+/*
+* Copyright 2019 elementary, Inc. (https://elementary.io)
+*
+* This program is free software; you can redistribute it and/or
+* modify it under the terms of the GNU General Public
+* License as published by the Free Software Foundation; either
+* version 3 of the License, or (at your option) any later version.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+* General Public License for more details.
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*
+*/
+
+.due-date {
+    border-radius: 12px;
+    background: lightgray;
+}
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -70,7 +70,7 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
                 due.second
             );
 
-            var due_label = new Gtk.Label (due_datetime.format ("%x %X"));
+            var due_label = new Gtk.Label (Granite.DateTime.get_relative_datetime (due_datetime));
             due_label.wrap = true;
             due_label.xalign = 0;
             due_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -21,10 +21,17 @@
 public class Tasks.TaskRow : Gtk.ListBoxRow {
     public unowned ICal.Component component { get; construct; }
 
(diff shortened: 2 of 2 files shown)


danrabbit:
I wonder if it makes sense to use https://valadoc.org/granite/Granite.DateTime.get_relative_datetime.html for this?

marbetschar:
Changed the code in order to test `Granite.DateTime.get_relative_datetime`. Not quite sure if we should use it in this context though. Here's how it behaves:

- If the due date is within the same week, it outputs the weekday in short format (Mon, Tue, ...)
- If the due date is within the same year, it outputs month name in short format and date (Oct 11, Nov 24, ...)
- Only if the date is not within the same year, it outputs the full date

That said, I'm not sure if this behaviour is what we want. What do you think?

danrabbit:
@marbetschar Do you feel like it's necessary to output the full date always? I think personally for a task item that I like the more natual-language feel of relative date/times

marbetschar:
@danrabbit I must admit, I got used to the full date approach because of Apple Reminders - so I'm definetely biased here. However I can see a potential benefit of the more natural-language feel. But sometimes this natural-language thing sometimes make me think that the date output is part of the task description instead of a standalone item with additional information - and I end up reading the same information twice till I figure out why.

One more thing to note is, that the natural-language output does not seem to print the time information if the task is not due on the same day. That's something which I think makes planning a bit harder.

Tu sum up: I'm really not sure what the best approach is here.

alainm23:
@marbetschar You could use: [Granite.DateTime.get_default_date_format](https://valadoc.org/granite/Granite.DateTime.get_default_date_format.html)

And make a validation, if the year of the task is the same as the year of the current date, do not show the year.

```
public string get_default_date_format_from_date (GLib.DateTime date) {
        var now = new GLib.DateTime.now_local ();

        if (date.get_year () == now.get_year ()) {
            return date.format (Granite.DateTime.get_default_date_format (false, true, false));
        } else {
            return date.format (Granite.DateTime.get_default_date_format (false, true, true));
        }
 }
```

danrabbit:
@marbetschar what do you think of putting the due date/time in the top/right instead of below the title? This would be in line with where other apps (like mail) display this information:

Screenshot from 2019-10-21 16 19 41@2x
Screenshot from 2019-10-21 16 20 02@2x



Interestingly, it looks like Things puts that info before the title:

Screenshot from 2019-10-21 16 18 21@2x

Would putting the full date/time in a tooltip be an acceptable compromise?

marbetschar:
@danrabbit I feel like the due date is an important information if set. Therefore hiding it in a tooltip or move it all across to the opposite side of the summary seems counter intuitive - at least for me.

On the other side I actually quite like the Things approach: Placing it in front of the summary and set it apart graphically. Depending how much other info is shown before the summary (Prio, ...?), this seems to be a good approach to me - which is capable using natural-language without causing confusion.

In addition, if the date is styled like this we can still put it on a dedicated line without my initial concerns I think.

I‘ll give it a try as soon as possible.

marbetschar:
@danrabbit I did a first rough draft and it looks quite promising. What do you think?

Screenshot from 2019-10-22 15 52 32@2x

Was not able to get the styling via CSS to work (using the newly created `taskrow_provider`), maybe you have an idea why? The code in question is [this one](https://github.com/elementary/tasks/blob/27b66edc6567a6c96f1f81f85287addcc242c640/src/Widgets/TaskRow.vala#L87).

danrabbit:
I tried it in front of the description but I really don't like how this makes things aligned differently:

Screenshot from 2019-10-22 08 33 00@2x

Instead I think putting it next to the description seems to work better:

Screenshot from 2019-10-22 08 50 33@2x

https://github.com/elementary/tasks/pull/51

PR #55, opened 2019-10-22
Accentuate checkbox

marbetschar:
Show checkbox in color of list.


src/Widgets/TaskRow.vala
```diff
@@ -40,8 +41,8 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         var check = new Gtk.CheckButton ();
         check.active = completed;
         check.margin_top = 2;
-        check.sensitive = false;
```
danrabbit: Can we keep this insensitive while it doesn't actually function?

marbetschar:
Sure thing!

Issue #45, opened 2019-10-19
Set tasks status to complete or incomplete by changing its checkbox state


marbetschar:
What the title says.

marbetschar:
Made a first attempt but toggling the state by using the `CheckButton`'s signal and executing `component.set_status` throws a `SegmentationFault`. Any idea why?

The code in question is in the toggle-state branch and does the following:

```
...
construct {
  completed = component.get_status () == ICal.PropertyStatus.COMPLETED;

  var check = new Gtk.CheckButton ();
  check.active = completed;
  check.toggled.connect (() => {
    completed = check.active;

    // this throws a Segmentation fault. Why?
    component.set_status(completed ? ICal.PropertyStatus.COMPLETED : ICal.PropertyStatus.NONE);
  });
...
```

marbetschar:
**UPDATE**: Obviously the previous code was trying to set the wrong status. This is fixed now in the `toggle-state` branch. However, there is still a `SegmentationFault`. It seems to always occur when we try to set `ICal.PropertyStatus.NONE`.

Any ideas anyone?

marbetschar:
@danrabbit checked the [`CalendarModel`](https://github.com/elementary/calendar/blob/master/core/Model/CalendarModel.vala) and tried to adapt. For some reason the `SegmentationFault` seems to be gone - no idea why.

But the `ICal.Component` does seem to loose it's data when I pass it through a Signal for saving. Therefore the `ECal.Client` fails to save, claiming it can't find any matching entity. Any idea what's wrong?

The code in question is located here:
https://github.com/elementary/tasks/blob/toggle-state/src/Widgets/TaskRow.vala#L38

marbetschar:
It should be possible to toggle the task status by:

- [ ] Click the checkbox
- [ ] Select the task row and hit SPACE

marbetschar:
@danrabbit or @tintou any idea why `component.get_uid ()` is empty in `on_component_update` of `ListView`?

Trying to pass the `ICal.Component` for updating like this:
1. [Firing `TaskRow.component_update` from bind `CheckButton.toggled`](https://github.com/elementary/tasks/blob/133dff3daf3f4818dd0181879d11c169e5914d69/src/Widgets/TaskRow.vala#L44)
    - `component.get_uid ()` is set
[...]

PR #61, opened 2019-12-14
Persist task changes to ECal backend

### LINKED ISSUE #60: Save task changes
What the title says: We need to persist any changes the user made to a task.

marbetschar:
This pull requests adds the possibility to actually modify a task and get the changes saved into the backend by using libecal.

**EDIT:** Fixes #60.
**EDIT 2:** Fixes https://github.com/elementary/tasks/issues/45 as well.

### DIFF
--- a/src/ListView.vala
+++ b/src/ListView.vala
@@ -200,7 +200,7 @@ public class Tasks.ListView : Gtk.Grid {
         } while (task_row != null);
     }
 
-    #if E_CAL_2_0
+#if E_CAL_2_0
     private void on_objects_removed (E.Source source, ECal.Client client, SList<ECal.ComponentId?> cids) {
 #else
     private void on_objects_removed (E.Source source, ECal.Client client, SList<weak ECal.ComponentId?> cids) {
--- a/src/ListView.vala
+++ b/src/ListView.vala
@@ -207,7 +207,7 @@ public class Tasks.ListView : Gtk.Grid {
 #endif
         debug (@"Received $(cids.length()) removed task(s) for source '%s'", source.dup_display_name ());
 
-        Tasks.TaskRow task_row = null;
+        unowned Tasks.TaskRow? task_row = null;
         var row_index = 0;
         do {
             task_row = (Tasks.TaskRow) task_list.get_row_at_index (row_index);
--- a/src/ListView.vala
+++ b/src/ListView.vala
@@ -135,12 +135,15 @@ public class Tasks.ListView : Gtk.Grid {
             }
         });
 
-        tasks_added (source, added_tasks.read_only_view);
+        tasks_added (client, source, added_tasks.read_only_view);
     }
 
-    private void tasks_added (E.Source source, Gee.Collection<ECal.Component> tasks) {
+    private void tasks_added (ECal.Client client, E.Source source, Gee.Collection<ECal.Component> tasks) {
         tasks.foreach ((task) => {
             var task_row = new Tasks.TaskRow (source, task);
+            task_row.changed_task.connect((task) => {
+                update_task (client, task, ECal.ObjModType.ALL);
+            });
             task_list.add (task_row);
             return true;
         });
@@ -159,5 +162,22 @@ public class Tasks.ListView : Gtk.Grid {
         }
 
         return 0;
+    }
+
+    public void update_task (ECal.Client client, ECal.Component task, ECal.ObjModType mod_type) {
+        unowned ICal.Component comp = task.get_icalcomponent ();
+        debug (@"Updating task '$(comp.get_uid())' [mod_type=$(mod_type)]");
+
+#if E_CAL_2_0
+        client.modify_object.begin (comp, mod_type, ECal.OperationFlags.NONE, null, (obj, results) => {
+#else
+        client.modify_object.begin (comp, mod_type, null, (obj, results) => {
+#endif
+            try {
+                client.modify_object.end (results);
+            } catch (Error e) {
+                warning (e.message);
+            }
+        });
     }
 }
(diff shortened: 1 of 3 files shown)


marbetschar:
@tintou in this pull request I also tried to show/hide the `description_grid` depending on the `ECal.Component task`'s data (the grid contains the `due_label` and `description_label`).

Unfortunately after first load, the `description_grid` is always shown. It only gets hidden after you modify a task the first time.

I have no clue why, because the very same code (`update_request (ECal.Component task)` in `TaskRow`) is responsible for it. Maybe you have an idea how to fix that?

src/Widgets/TaskRow.vala
```diff
@@ -67,25 +66,97 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+            if (task == null) {
+                return;
+            }
+            task.get_icalcomponent ().set_status (check.active ? ICal.PropertyStatus.COMPLETED : ICal.PropertyStatus.NONE);
+            task_changed (task);
+        });
+
+        key_release_event.connect ((event) => {
+            // SPACE == 32
+            if (event.keyval == 32) {
```
danrabbit: There's `Gdk.Key` for this :) 
```suggestion
            if (event.keyval == Gdk.Key.space) {
```

src/Widgets/TaskRow.vala
```diff
@@ -67,25 +66,97 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+                check.active = !check.active;
+            }
+        });
+    }
+
+    public void update_request (ECal.Component task) {
+        this.task = task;
```
danrabbit: Instead of doing things this way around, I think I would probably have task be a `public { get; construct set; }` and then add a `notify["task"].connect ()`. I think it feels more consistent to set object properties using properties instead of methods :)

src/Widgets/TaskRow.vala
```diff
@@ -67,25 +66,97 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+        if (completed) {
+            summary_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+            description_grid.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+        } else {
+            summary_label.get_style_context ().remove_class (Gtk.STYLE_CLASS_DIM_LABEL);
+            description_grid.get_style_context ().remove_class (Gtk.STYLE_CLASS_DIM_LABEL);
+        }
+
+        if ( ical_task.get_due ().is_null_time () ) {
+            due_label.visible = false;
```
danrabbit: visible property has the problem that `show_all ()` overrides it. In general it's better to just use [Gtk.Revealer](https://valadoc.org/gtk+-3.0/Gtk.Revealer.html). Especially because you get nice transitions and we want that anyways

src/Widgets/TaskRow.vala
marbetschar: good catch! I really wanted it to be this way, but wasn't aware that we can combine `construct set;` - I assumed it's an either/or. Will update the code to reflect this!

src/Widgets/TaskRow.vala
marbetschar: Understood. Will change the Code accordingly. Stumbled across Gtk.Revealer before and was wondering what it might be doing anyways ;)

marbetschar: Implemented the suggested change. However, it seems I still need to set the `due_label_revealer`'s visibility to `false` after it has hidden it's child in order to free the space and allow the description to span over the whole row.

I hope I did this the way suggested. If there's a better way to do this I happily apply it.

marbetschar:
@danrabbit you're very welcome!! Thanks for your review. Feel free to add additional comments if there are further spots which need to be improved.

src/ListView.vala
```diff
@@ -127,26 +129,104 @@ public class Tasks.ListView : Gtk.Grid {
...
+                        task_row.task = task;
+                        break;
+                    }
+                }
+            }
+            row_index++;
+        } while (task_row != null);
+    }
+
+    #if E_CAL_2_0
```
tintou:
```suggestion
#if E_CAL_2_0
```

src/ListView.vala
```diff
@@ -127,26 +129,104 @@ public class Tasks.ListView : Gtk.Grid {
...
+    private void on_objects_removed (E.Source source, ECal.Client client, SList<ECal.ComponentId?> cids) {
+#else
+    private void on_objects_removed (E.Source source, ECal.Client client, SList<weak ECal.ComponentId?> cids) {
+#endif
+        debug (@"Received $(cids.length()) removed task(s) for source '%s'", source.dup_display_name ());
+
+        Tasks.TaskRow task_row = null;
```
tintou:
```suggestion
        unowned Tasks.TaskRow? task_row = null;
```

src/ListView.vala
```diff
@@ -127,26 +129,104 @@ public class Tasks.ListView : Gtk.Grid {
...
+        var row_index = 0;
+        do {
+            task_row = (Tasks.TaskRow) task_list.get_row_at_index (row_index);
+
+            if (task_row != null) {
+                foreach (ECal.ComponentId cid in cids) {
```
tintou:
```suggestion
                foreach (unowned ECal.ComponentId cid in cids) {
```

src/ListView.vala
```diff
@@ -160,4 +240,35 @@ public class Tasks.ListView : Gtk.Grid {
...
+#else
+        client.modify_object.begin (comp, mod_type, null, (obj, results) => {
+#endif
+            try {
+                client.modify_object.end (results);
+                SList<ECal.Component> ecal_tasks;
+                client.get_objects_for_uid_sync (comp.get_uid (), out ecal_tasks, null);
+
+                var ical_tasks = new SList<unowned ICal.Component> ();
+                foreach (ECal.Component ecal_task in ecal_tasks) {
```
tintou:
```suggestion
                foreach (unowned ECal.Component ecal_task in ecal_tasks) {
```

marbetschar:
Fixed as suggested by @tintou. Is the `unowned` generally a good practice, or is in this case advisable just because we deal with a `SList<weak ...` respectively `SList<unowned` ...?

PR #52, opened 2019-10-22
Bold summary label

marbetschar:
Made the summary label bold, since the due date does grab a lot of attention otherwise:

Screenshot from 2019-10-22 19 04 25@2x

PR #63, opened 2019-12-18
TaskRow: Make task descriptions small

danrabbit:
Alternative to #52


cassidyjames:
@danrabbit does it make more sense to use CSS or pango `<small>` here?

danrabbit:
@cassidyjames Yeah can do. I just wanted to make sure to match between the description text and due label. Maybe I should change them both to Pango?

I'm wondering about if we should have a CSS class for "SMALL_LABEL". Using Pango seems to make things more complicated when the labels needs to change

cassidyjames:
@danrabbit yeah a constant in Granite could make sense, and matches the convention with dim label. We should probably make it match whatever Pango does to start for consistency, but once we move things over to it, we'd have more freedom with the styling.

src/Widgets/TaskRow.vala
```diff
@@ -86,7 +86,12 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         description_label.xalign = 0;
         description_label.lines = 1;
         description_label.ellipsize = Pango.EllipsizeMode.END;
-        description_label.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
+        description_label.use_markup = true;
+
+        unowned Gtk.StyleContext description_label_context = description_label.get_style_context ();
+        description_label_context.add_provider (taskrow_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
+        description_label_context.add_class ("description");
```
cassidyjames:
```suggestion
        // NOTE: This should change to using the constant after a release of Granite with elementary/granite#367
        description_label_context.add_class ("small-label");
```

jeremypw:
@danrabbit This was approved a while back but now has several conflicts - is it still relevant?

danrabbit:
Yeah I'm just gonna close this and we can revisit it at another time There are other styling things that need to be addressed before initial release

PR #65, opened 2019-12-18
ListView: Add a placeholder for empty task lists

danrabbit:


src/ListView.vala
```diff
@@ -46,8 +46,16 @@ public class Tasks.ListView : Gtk.Grid {
         settings_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
         settings_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
 
+        var placeholder = new Gtk.Label (_("No Reminders"));
```
cassidyjames: Not all tasks are/have reminders

```suggestion
        var placeholder = new Gtk.Label (_("No Tasks"));
```

src/ListView.vala
```diff
@@ -46,8 +46,16 @@ public class Tasks.ListView : Gtk.Grid {
...
+        placeholder.show ();
```
cassidyjames: It looks way too low visually, would this be awful?

```suggestion
        placeholder.yalign = 0.33;
        placeholder.show ();
```

Screenshot from 2019-12-18 12 23 30

danrabbit: I think this will resolve itself when there's UI for adding tasks at the bottom of the view. I'd be inclined to hold off on this for now

PR #66, opened 2019-12-18
TaskSettingsPopover: Add a frame and margins around description

danrabbit:
Fixes a visual issue with the popover arrow:

## BEFORE
Screenshot from 2019-12-18 09 04 10@2x

## AFTER
Screenshot from 2019-12-18 09 02 42@2x


alainm23:
Could the datepicker be aligned? I feel that it is more inside than textarea and entry.

danrabbit:
@alainm23 Yeah I'd like to follow up on this in another branch. It seems to be its own widget for some reason and I don't see a need for that

alainm23:
Great, good eyes 👀️.

Issue #54, opened 2019-10-22
Add a new task


danrabbit:
We need a method to add a new task to the selected task list

alainm23:
### My design:

```
------------------------------------------------------------------ -> Gtk.Separator
[ x ] [Task name -> Gtk.Entry]
      [Task note -> Gtk.TextVIew ]
       
      [ datetime buttom ]                         [Cancel]  [Save]
------------------------------------------------------------------ -> Gtk.Separator
[Save button] -> Flat -> text color depending on the color of the list
[Cancel button] -> Flat
```
How should this work?
- Could we add it as a row in the task list?
- At the top or end of the task list as a Gtk.Revealer

danrabbit:
Something I think seemed pretty smooth was the approach of Agenda to just have a text entry at the bottom of the view and you could type something and hit enter and then it creates the task. That kept things predictable and made it easy to add a bunch of tasks at once

marbetschar:
I agree with @danrabbit: I also quite like the approach of having a text entry at the bottom of the list. For now we may simply start with making the task summary editable within the list view (similar to how its done for the list title, without the need of double click).

If done like this, we can simply add additional, but empty row at the end of the list which creates a new task.

lainsce:
I have an idea for this:

Instead of making a new row just for the new task entry, we shift this action to a floating action button in the corner, which when clicked, opens up a `New task…` popover. This button would be colored with the task list's color, and sized 32x32 on LoDPi / 64x64 on HiDPI, for easy discoverability.

Design:
![granite_fab_mockup](https://user-images.githubusercontent.com/4886639/72587739-e423e000-38d4-11ea-9fc5-8d83ebe76d01.png)

danrabbit:
Fixed in #89

Issue #69, opened 2019-12-18
UX improvements to edit tasks.


alainm23:
1. Is it intuitive to double click on the task to edit it?
- Double clicking is fine, but we can add a button at the end of the task so that the edition is also activated (and have the Popover shown from this button)

2. Focus for the Gtk.Entry of the task name.
- When the popover is shown we can make the name of the task focus, this with the intention of making the edition faster.

danrabbit:
I'm not a big fan of the double click personally. We don't normally use Double click other places, so I feel like this could at least be a single click. But in general, I'm not sure a popover is the right approach. It seems like more inline editing would be better. Planner seems to handle this a lot smoother

marbetschar:
I just used the popover to get started, since I'm not too familiar yet with the possibilities of Gtk - and use Apple Reminders a lot. But the UX there is quite old fashioned.

That said +1 from me to get rid of the popover ;)

Inline editing seems much better. Also the way Planner does it is super cool!  The only thing there is, I'm not 100% convinced of the "transform to card" approach. It is maybe even smoother if we just slide the road down on selection to show the form - and preselect the summary entry.

marbetschar:
```
|////////////////////////////////////////////|
|                                            |
| My Task List                           [:] |
|                                            |
|--------------------------------------------|
|                                            |
| [+] [___________________________________]  |    ← (1)
|                                            |
|--------------------------------------------|
|                                            |
|  ☐ Design needs improvement                |
|    The task app should be intiutive a...   |
|                                            |
| 🢒 ☐ Design Task App   ⬉                    |    ← (2)
|    (in 40min) The task app should be ...   |
|                                            |
|============================================|
| 🢓 ☐ The Edit UI                            |    ← (3)
|  This is the very long description, which  |
|  can span over multiple lines.             |
|                                            |
|  Lorem ipsum ... sit amet.                 |
|--------------------------------------------|
|  Schedule  ☐🗹   [ 1st Jan 📅][ 8:30 pm 🕑] |
|--------------------------------------------|
|  [ Delete ]               [Cancel] [Save]  |
|============================================|
|                                            |
|  ☐ Design needs improvement                |
|                                            |
|  ☐ Design needs improvement                |
|    The task app should be intiutive a...   |
|                                            |
|--------------------------------------------|
|                                            |
|////////////////////////////////////////////|
```
A first attempt to improve the task workflow:
[...]

PR #76, opened 2019-12-21
Made inline editing of task summary possible along with some keyboard…

marbetschar:
… shortcuts (ESC, Enter, Space).

This PR contains the first steps towards #54.


marbetschar:
@danrabbit any suggested improvements to get this PR merged? Looking forward to tackle #69 based upon this work.

src/Widgets/TaskRow.vala
```diff
@@ -130,8 +118,49 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+                    if (has_focus) {
+                        show_popover ();
+                    } else {
+                        grab_focus ();
+                    }
+                    break;
+
+                case Gdk.Key.BackSpace:
+                    if (has_focus) {
+                        // TODO: Delete Task
```
danrabbit: In general, I would prefer if we didn't merge any dead code. We can use GitHub issues to describe ToDo's :)

marbetschar: If I got this straight you mean I should remove the `case Gdk.Key.BackSpace: ...` from the PR, right?

marbetschar: Dropped the case for BackSpace.

marbetschar:
@danrabbit if you find some time it would be awesome to get this merged. Will tackle inline task editing next - based upon this work (#69).

PR #78, opened 2020-01-26
Inline editing

marbetschar:
Added inline editing possibilities for a given task:

Screenshot from 2020-01-26 16-09-37

This branch is based upon the work done in `inline-edit-task-summary`.


alainm23:
Maybe, We can add a Gtk.Separator to know where the task edition starts or ends.

marbetschar:
Here you go: Separators & selected style.

Screenshot from 2020-01-27 08-40-06

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+
+    public TaskFormRevealer (ECal.Component task) {
+        Object (task: task);
+    }
+
+    construct {
+        due_switch = new Gtk.Switch ();
+        due_switch.valign = Gtk.Align.CENTER;
+
+        var due_label = new Gtk.Label (_("Schedule") + ":");
```
tintou:
```suggestion
        var due_label = new Gtk.Label (_("Schedule:"));
```
The i18n requires full string. For instance we need an unbreakable space before colons in french

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+        due_timepicker = new Granite.Widgets.TimePicker ();
+
+        var due_datetimepicker = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 12);
+        due_datetimepicker.add (due_label);
+        due_datetimepicker.add (due_switch);
+        due_datetimepicker.add (due_datepicker);
+        due_datetimepicker.add (due_timepicker);
+
+        var description_textview = new Gtk.TextView ();
+        description_textview.left_margin = description_textview.right_margin = 12;
```
tintou:
```suggestion
        description_textview.margin = 12;
```
is equivalent to giving everything a margin

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+            due_datepicker.date = due_timepicker.time = null;
+        } else {
+            var due_date_time = Util.ical_to_date_time (ical_task.get_due ());
+            due_datepicker.date = due_timepicker.time = due_date_time;
+
+            due_switch.active = true;
+        }
+
+        if (ical_task.get_description () != null) {
+            description_textbuffer.text = ical_task.get_description ().strip ();
```
tintou:
```suggestion
            description_textbuffer.text = ical_task.get_description ();
```
If one want to put a leading or ending space they are free to do it

src/Widgets/TaskDetailRevealer.vala
```diff
@@ -0,0 +1,143 @@
...
+
+                due_label_revealer.reveal_child = true;
+            }
+
+            if (ical_task.get_description () == null) {
+                description_label_revealer.reveal_child = false;
+
+            } else {
+                var description = ical_task.get_description ();
+                description = description.replace ("\r", "").strip ();
```
tintou: I don't like what's happening to the description, it should remain unprocessed

marbetschar: The processing of the task description is becaues in "view-mode" the description is rendered to display on a single line as discussed [here](https://github.com/elementary/tasks/issues/35#issuecomment-543216958).

This idea origins of how the Apple Reminder rendered the task description prior to macOS Catalina.

Is there a better way to do this?

tintou: I've never used an Apple computer so I might need to search for screenshots as I don't understand what's the desired effect here.

tintou: A cleaner way would be to have a utility function (in a Utils.vala file?) that would return a wrapped string.
It would probably be more efficient to use a GLib.Regex to find an replace `(^\s+|\s+$|\n|\s\s+)` with a space :slightly_smiling_face:

marbetschar: *haha* [originally wanted to use Regex](https://github.com/elementary/tasks/pull/39), but did not figure out how 🤭 thanks for the hint!

marbetschar: Done using GLib.Regex.

marbetschar:
Once everything is Ok, can you guys then also merge the `inline-edit-task-summary` into master? This pull request was made on top of the other branch and contains all changes also made there.

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+        due_timepicker = new Granite.Widgets.TimePicker ();
+
+        var due_datetimepicker = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 12);
+        due_datetimepicker.add (due_label);
+        due_datetimepicker.add (due_switch);
+        due_datetimepicker.add (due_datepicker);
+        due_datetimepicker.add (due_timepicker);
+
+        var description_textview = new Gtk.TextView ();
+        description_textview.margin = 12;
```
danrabbit: You probably want `border_width` instead of `margin` since `border_width` is internal to the text view and you won't get that grey around the outside

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+
+        var button_box = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL);
+        button_box.baseline_position = Gtk.BaselinePosition.CENTER;
+        button_box.set_layout (Gtk.ButtonBoxStyle.END);
+
+        button_box.add (delete_button);
+        button_box.set_child_secondary (delete_button, true);
+        button_box.set_child_non_homogeneous (delete_button, true);
+        button_box.add (cancel_button);
+        button_box.set_child_non_homogeneous (cancel_button, true);
```
danrabbit: Usually we do want these buttons to be homogeneous. Is there a reason you set them to not be?

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+        description_scrolled_window.hscrollbar_policy = Gtk.PolicyType.EXTERNAL;
+        description_scrolled_window.height_request = 140;
+        description_scrolled_window.add (description_textview);
+
+        var description_frame = new Gtk.Frame (null);
+        description_frame.margin_bottom = 0;
+        description_frame.add (description_scrolled_window);
+
+        var delete_button = new Gtk.Button ();
+        delete_button.sensitive = false;
```
danrabbit: Instead of hardcoding this, I imagine this should be tied to some deletable property?

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+        description_textbuffer = new Gtk.TextBuffer (null);
+        description_textview.set_buffer (description_textbuffer);
+
+        var description_scrolled_window = new Gtk.ScrolledWindow (null, null);
+        description_scrolled_window.hscrollbar_policy = Gtk.PolicyType.EXTERNAL;
+        description_scrolled_window.height_request = 140;
+        description_scrolled_window.add (description_textview);
+
+        var description_frame = new Gtk.Frame (null);
+        description_frame.margin_bottom = 0;
```
danrabbit: It doesn't look like you add margin to the frame anywhere else. It should be 0 by default already

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+        description_textview.margin = 12;
+        description_textview.top_margin = description_textview.bottom_margin = 12;
+        description_textview.set_wrap_mode (Gtk.WrapMode.WORD_CHAR);
+        description_textview.accepts_tab = false;
+
+        description_textbuffer = new Gtk.TextBuffer (null);
+        description_textview.set_buffer (description_textbuffer);
+
+        var description_scrolled_window = new Gtk.ScrolledWindow (null, null);
+        description_scrolled_window.hscrollbar_policy = Gtk.PolicyType.EXTERNAL;
```
danrabbit: Won't have to do this if you use `border_width` for the text view

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+
+    construct {
+        due_switch = new Gtk.Switch ();
+        due_switch.valign = Gtk.Align.CENTER;
+
+        var due_label = new Gtk.Label (_("Schedule:"));
+        due_datepicker = new Granite.Widgets.DatePicker ();
+        due_timepicker = new Granite.Widgets.TimePicker ();
+
+        var due_datetimepicker = new Gtk.Box (Gtk.Orientation.HORIZONTAL, 12);
```
danrabbit: Probably not necessary to add a box here when we can use Grid :)

src/Widgets/TaskFormRevealer.vala
```diff
@@ -0,0 +1,173 @@
...
+        button_box.set_layout (Gtk.ButtonBoxStyle.END);
+
+        button_box.add (delete_button);
+        button_box.set_child_secondary (delete_button, true);
+        button_box.set_child_non_homogeneous (delete_button, true);
+        button_box.add (cancel_button);
+        button_box.set_child_non_homogeneous (cancel_button, true);
+        button_box.add (save_button);
+
+        var box = new Gtk.Box (Gtk.Orientation.VERTICAL, 12);
```
danrabbit: Instead of two boxes, I would recommend Gtk.Grid

src/Widgets/TaskFormRevealer.vala
marbetschar: Fixed.

src/Widgets/TaskFormRevealer.vala
marbetschar: Changed the delete button to be homogeneous, while keeping the cancel button as not homogeneous. The reason for this is, I figured the alignment of the cancel button is better, when its not homogeneous:

Screenshot from 2020-02-01 13-25-44

If the cancel button is homogeneous too, it feels out of place:

Screenshot from 2020-02-01 13-25-11

src/Widgets/TaskFormRevealer.vala
marbetschar: Exactly. Because delete functionality is not implemented yet, I made the delete button non-sensitive.

src/Widgets/TaskFormRevealer.vala
marbetschar: Fixed!

src/Widgets/TaskFormRevealer.vala
marbetschar: Dropped the `Gtk.ScrolledWindow` completely and added a `description_textview.height_request = 140`.

We now have a `TextView` which adapts its height to its content.

marbetschar:
> ```
> ../src/Util.vala:38.16-38.85: warning: unhandled error `GLib.RegexError'
>         return Tasks.Util.line_break_to_space_regex.replace (str, str.length, 0, " ");
> ```

**Good catch! I added a try/catch surrounding the return statement:**

https://github.com/elementary/tasks/blob/9fabf5ff4be30c8b4dc92b01804929766a51299f/src/Util.vala#L38-L44

marbetschar:
> I get a ton of criticals when opening my shopping list. Looks like it isn't handled when things are null:
> 
> Screenshot from 2020-01-31 09 42 34@2x

**Seems like the `Granite.DateTime` Widget does not like `null` values, so I set it's value to `now_local` if there isn't any value provided by the task itself - but UX-wise I'm not sure if it's even better to hide the widget fi there is no value. What's the preferred approach here?**

https://github.com/elementary/tasks/blob/9fabf5ff4be30c8b4dc92b01804929766a51299f/src/Widgets/TaskFormRevealer.vala#L145

src/Widgets/TaskFormRevealer.vala
marbetschar: Tried to get use one `Gtk.Grid` instead of the two `Gtk.Box`, but I was not able to rebuild the interface using the Grid: The due fields are then spoiled over the whole width as soon as I set `description_frame.vexpand = true`. Additionally setting `grid.column_homogeneous = false` and the due field's `.halign =  Gtk.Align.START` did not help either.

Whats the approach to recreate the given interface with only one `Gtk.Grid`? And is it in general advised to use `Gtk.Grid` over nesting `Gtk.Boxes`? What are the drawbacks of nesting Widgets?

At the moment I feel like nesting boxes is a lot easier to wrap ones head around - but maybe that's just me, lacking the experience using `Gtk.Grid`.

src/Widgets/TaskFormRevealer.vala
marbetschar: See comment below, regarding wrangling with `Gtk.Grid` to rebuild the interface.

src/Widgets/TaskFormRevealer.vala
marbetschar: @danrabbit how shall we proceed with the `Gtk.Grid` / `Gtk.Box` thingy?

danrabbit:
Ah you know what, let's merge this into that other branch and continue reviewing there

PR #79, opened 2020-01-26
Add support for libecal-2.0

danrabbit:
for focal builds


tintou:
I'm not on focal right now but I'm sure that there is more work to do to make sure that it works for both

PR #80, opened 2020-02-01
Rename gschema.xml to io.elementary.tasks.gschema.xml

RavetcoFX:
This fixes a segfault in not finding the schema


danrabbit:
The meson build file renames this when it installs. This breaks the build system and won’t solve the error you’re getting. You need to install first and then you can run from the build directory

PR #89, opened 2020-03-04
Add task

marbetschar:
This Pull Request adds the ability to create a new task. To achieve this, an empty task row is displayed at the top.

Known issue: The new task gets sucessfully created. However, for some reason the task list does not get refreshed. The user needs to navigate to another list and come back to the current one in order to see the newly created task. See the following GIF:

New Task

### DIFF
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -68,7 +68,7 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
     }
 
     construct {
-        created = Tasks.Util.calcomponent_created (task);
+        created = calcomponent_created (task);
 
         icon = new Gtk.Image.from_icon_name ("list-add-symbolic", Gtk.IconSize.MENU);
         icon.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
@@ -442,5 +442,17 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
             destroy ();
             return GLib.Source.REMOVE;
         });
+    }
+
+    /*
+     * Returns whether or not a ECal.Component was created in the backend EDS.
+     */
+    private bool calcomponent_created (ECal.Component comp) {
+        if (comp == null) {
+            return false;
+        }
+        ICal.Time created;
+        comp.get_created (out created);
+        return !created.is_null_time ();
     }
 }
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -197,7 +197,7 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
 
         if (icon != null) {
             grid.attach (icon, 0, 0);
-        } else if(check != null) {
+        } else if (check != null) {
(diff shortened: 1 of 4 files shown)


src/Widgets/TaskRow.vala
```diff
@@ -122,23 +148,28 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
         var button_box = new Gtk.ButtonBox (Gtk.Orientation.HORIZONTAL);
         button_box.baseline_position = Gtk.BaselinePosition.CENTER;
         button_box.margin_top = 12;
         button_box.spacing = 6;
         button_box.set_layout (Gtk.ButtonBoxStyle.END);
-        button_box.add (delete_button);
+        if (delete_button != null) {
+            button_box.add (delete_button);
+        }
         button_box.set_child_secondary (delete_button, true);
```
danrabbit:
```suggestion
            button_box.add (delete_button);
            button_box.set_child_secondary (delete_button, true);
        }
```

src/Widgets/TaskRow.vala
```diff
@@ -19,12 +19,19 @@
 */
 
 public class Tasks.TaskRow : Gtk.ListBoxRow {
+
+    public enum EditMode {
```
danrabbit: Does this really need to be an enum? It seems like a bool would cover it

src/Widgets/TaskRow.vala
```diff
@@ -122,23 +148,28 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
+            delete_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+        }
 
         var cancel_button = new Gtk.Button ();
         cancel_button.label = _("Cancel");
 
         var save_button = new Gtk.Button ();
-        save_button.label = _("Save Changes");
         save_button.get_style_context ().add_class (Gtk.STYLE_CLASS_SUGGESTED_ACTION);
+        save_button.label = _(edit_mode == EditMode.ADD ? "Add Task" : "Save Changes");
```
danrabbit:
```suggestion
        save_button.label = edit_mode == EditMode.ADD ? _("Add Task") : _("Save Changes");
```

src/Widgets/TaskRow.vala
marbetschar: Enum is not mandatory, just thought it’s more expressive than a bool.

My personal favorite would be to detect this state based on the ECal.Component itself (some kind of is_new logic). But I did not figure out how.

marbetschar:
Packing the check and icon into a stack is a good idea.

Regarding the other stuff: do you have an idea how we can figure out if a ECal.Component was already added or not? We can then depend the other logic based on that - which makes is a bit more UI agnostic.

marbetschar:
In terms of the task not automatically been shown on the list, that seems to be related tohttps://github.com/elementary/tasks/issues/90 and will be tackled in this dedicated issue.

@danrabbit worked through your feedback and I think the code should be fine now. Can you give this another look so we can get this merged?

PR #98, opened 2020-03-28
Scheduled list

### LINKED ISSUE #46: Add "planned" list
## Prerequisites

## Feature
Right now its hard to see if I'm acting according to my created plan. That is, I can't see which tasks are planned to be done next and if there are any tasks which are overdue - regardless of the list they are in.

We should consider adding a new, dynamically created list which shows all "planned" tasks. Where "planned" is defined as tasks which either one of the following is true:

- due date is set to the past, but they are still incomplete
- due date is set to the future

**Existing work**
The querying routine is already in place, so we might just want to copy and adapt it accordingly.

**Additional context**
We need to define whether we want to show such a list per origin (Nextcloud, iCloud, ...) or just one consolidated list across all origins. In my opinion the consolidated list is preferable because it makes planning easier for the user.

marbetschar:
**ready for review**

This PR fixes #46 

Screenshot from 2020-03-28 12-34-23


marbetschar:
Blocked until a solution is found for #99

marbetschar:
This PR is based on the work done in #100 and displays a first version of the Scheduled list.

marbetschar:
Screenshot from 2020-04-13 22-14-53

marbetschar:
Would be great to add relative dates as group headers to the Scheduled list view as described here: https://github.com/elementary/granite/issues/381

marbetschar:
Added `get_relative_date` reference implementation which provides nice, localized date formatted strings:

Screenshot from 2020-04-17 08-38-07

And within the task list, we simply append the time:

Screenshot from 2020-04-17 08-37-59

danrabbit:
From Slack:

> This issue is that typically relative datetime is only dealing in dates that happen in the past. So if you open Mail and you see an email from "Mon" you know it was last monday, the 13th, and not an email from next monday the 20th. Wouldn't make sense in context right?
>
> But in Tasks, if you're looking at all scheduled tasks you could have missed tasks from last monday and upcoming tasks next monday, so you would have two headers "Mon" where one is all past-due and the other is upcoming.
>
> I do like the simplicity of the day name for things like Mail, Notifications, etc. So I kinda don't want to lose that. I also think it's more useful to me if I have an upcoming event on "Mon" instead of "April 20" and then I have to figure out what day that means. I wonder if we should leave last monday as just "Mon" and only disambiguate future monday as "Mon, Apr 20".
>
> Hm now that I'm writing it out I wonder if that's backwards and I want future events to be "Mon" and past ones to be "Mon, Apr 13".  Or what if we just did "Last Mon" "Next Mon"


Since we don't have any other use cases right now except for Tasks, it might make more sense to implement this function in Tasks and refine it there and then we can come back and propose something to Granite if it turns out other apps have this issue as well

marbetschar:
Alright. That said, the first version of the Scheduled list is ready for review!

src/MainWindow.vala
```diff
@@ -51,7 +50,7 @@ public class Tasks.MainWindow : Gtk.ApplicationWindow {
         add_action_entries (ACTION_ENTRIES, this);
 
         foreach (var action in action_accelerators.get_keys ()) {
-            ((Gtk.Application) GLib.Application.get_default ()).set_accels_for_action (ACTION_PREFIX + action, action_accelerators[action].to_array ());
+            ((Gtk.Application) GLib.Application.get_default ()).set_accels_for_action (ACTION_PREFIX + action, action_accelerators[action].to_array ());  // vala-lint=line-length
```
danrabbit, line 53: I'd prefer not to add exclusions for line length. We can probably do something about this to make it not so long

marbetschar: Using two lines now for this.

marbetschar:
Fixed Warnings and now using the default `@colorAccent` in the scheduled list. In terms of the low contrast label, I guess this is because we are using the `EditableLabel` there which is set to non-sensitive - setting the accent color does not seem to have any effect:

https://github.com/elementary/tasks/blob/98b43b558a58ab9774d0fe3b7f842f37ce98f7fc/src/ListView.vala#L147

src/ListView.vala
```diff
@@ -84,55 +121,54 @@ public class Tasks.ListView : Gtk.Grid {
...
+        if (source == null) {
+            editable_title.sensitive = false;
+            editable_title.text = _("Scheduled");
+            settings_button_revealer.reveal_child = false;
+
+            Tasks.Application.set_task_color_from_string ("@colorAccent", editable_title);
+
+            task_list.@foreach ((row) => {
+                if (row is Tasks.TaskRow) {
+                    (row as Tasks.TaskRow).update_request ();
```
danrabbit, line 155: The compiler doesn't like this: `Access to possible 'null'. Perform a check or use an unsafe cast.`

Since we do check if the row is a type of `Tasks.TaskRow` we can probably cast like:
```suggestion
                    ((Tasks.TaskRow) row).update_request ();
```

src/ListView.vala
```diff
@@ -147,78 +183,93 @@ public class Tasks.ListView : Gtk.Grid {
...
+
+            if (comp_before.get_due ().compare_date_only (comp.get_due ()) == 0) {
+                return;
+            }
+        }
+
+        var due_date_time = Util.ical_to_date_time (comp.get_due ());
+        var header_label = new Granite.HeaderLabel (Tasks.Util.get_relative_date (due_date_time));
+        header_label.ellipsize = Pango.EllipsizeMode.MIDDLE;
+        header_label.margin_left = 4;
```
danrabbit, line 247: `margin_left` is deprecated. Also doesn't align with the editing card. Should be:
```suggestion
        header_label.margin_start = 6;
```

src/MainWindow.vala
```diff
@@ -174,12 +247,16 @@ public class Tasks.MainWindow : Gtk.ApplicationWindow {
 
         var header_label = new Granite.HeaderLabel (display_name);
         header_label.ellipsize = Pango.EllipsizeMode.MIDDLE;
+        header_label.margin_left = 4;
```
danrabbit, line 250: Same as above:
```suggestion
        header_label.margin_start = 6;
```

marbetschar:
Superseeded by #120

PR #100, opened 2020-03-30
Using task model to detach UI from EDS

### LINKED ISSUE #99: Abstraction between UI and EDS
During my attempt to create a scheduled task list I came across the situation where it is useful to untangle the EDS `E.SourceRegistry` and `E.Source` objects from UI interactions:

In order to retrieve the data from the eds backend, we need to initiate a `E.SourceRegistry` which in turn provides all available `E.Source` objects (aka "task lists").

**How it works now:** If we now want to display the tasks of each source in a dedicated list, we can simply pass the `E.Source` object to the UI list element and load the tasks from there. That's exactly how its done now.

**The downside of this approach:** If we want to display a list of scheduled tasks which are provided by more than one `E.Source` object, there is no easy way to integrate this using the current state of work. We need to start duplicating code and adapt it to allow more than one of those `E.Source` objects.

**That said:** it seems to be cleaner to refactor the way tasks are handled (loaded, changed, deleted) right now in order to support both scenarios using a UI agnostic approach. Similar to how it is done [in Planner](https://github.com/alainm23/planner/blob/b2c545003bebd541b01b2e4b14cf4df8f8a32e30/src/Services/Calendar/CalendarModel.vala) or [in Calendar](https://github.com/elementary/calendar/blob/master/core/Model/CalendarModel.vala).

marbetschar:
This draft attempts to fix #99 in order to allow more flexible querying possibilities to create views for retrieving tasks. Needed by #98.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,6 +34,7 @@ executable(
     'src/Widgets/SourceRow.vala',
     'src/Widgets/TaskRow.vala',
     'src/Widgets/ListSettingsPopover.vala',
+    'src/TaskModel.vala',
     dependencies: [
         dependency('glib-2.0'),
         dependency('gobject-2.0'),
--- a/src/Application.vala
+++ b/src/Application.vala
@@ -31,7 +31,7 @@ public class Tasks.Application : Gtk.Application {
 
     static construct {
         settings = new Settings ("io.elementary.tasks");
-        model = new Tasks.TaskModel ();
+        model = new Tasks.TaskModel();
     }
 
     protected override void activate () {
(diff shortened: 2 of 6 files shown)


marbetschar:
There is a working proof of concept which verifies this approach works for querying multiple sources and display all results in one single `ListView` in the `scheduled-list` branch. That said, this PR paves the way to create the Scheduled list:

https://github.com/elementary/tasks/blob/e2bfe452e55e9611a329db2e29885e02b75645be/src/MainWindow.vala#L154

PR #102, opened 2020-04-06
Made sure we always set the due date to the next full hour if not def…

### LINKED ISSUE #96: Set due date to next full hour
If the due date is activated by the user, we should set it's time to the next full hour instead of "now".

marbetschar:
**ready for review**

…ined otherwise. This fixes #96.


src/Widgets/TaskRow.vala
```diff
@@ -347,9 +347,13 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
             task_detail_revealer.reveal_child = false;
             task_detail_revealer.get_style_context ().remove_class (Gtk.STYLE_CLASS_DIM_LABEL);
 
+            var due_time = new DateTime.now_local ().add_hours (1);
```
danrabbit, line 350: I know it's a little nitpicky, but can we change this to `default_due_datetime`? I feel like It helps keep it clear when we're setting the picker that this isn't a datetime that we're getting from EDS and it's a datetime not just a time. Starting on line 374 we have:

```
                due_datepicker.date = due_timepicker.time = due_time;
            } else {
                var due_date_time = Util.ical_to_date_time (ical_task.get_due ());
                due_datepicker.date = due_timepicker.time = due_date_time;
```

so you can see how these look extremely similar

marbetschar: @danrabbit I love nitpicking ;) Makes total sense! Will rework this little feature, once we are fine with the new `Calendar.Store`.

marbetschar: renamed variable and moved its initialization outside of the conditional.

danrabbit (approved):
Seems reasonable :)

PR #103, opened 2020-04-08
Checkboxes update their color

### LINKED ISSUE #70: Check item color doesn't update
When changing the task list color, check items should update to match the new color

marbetschar:
Fixes #70:

Checkbox Color

### DIFF
--- a/data/TaskRow.css
+++ b/data/TaskRow.css
@@ -48,11 +48,6 @@ row.card {
     padding-top: 6px;
 }
 
-row check {
-    padding: 1px;
-    border-radius: 12px;
-}
-
 row entry.flat  {
     background: transparent;
     border-style: solid;
--- a/data/TaskRow.css
+++ b/data/TaskRow.css
@@ -48,6 +48,11 @@ row.card {
     padding-top: 6px;
 }
 
+row check {
+    padding: 1px;
+    border-radius: 12px;
+}
+
 row entry.flat  {
     background: transparent;
     border-style: solid;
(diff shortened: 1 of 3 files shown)


data/TaskRow.css
```diff
@@ -48,6 +48,11 @@ row.card {
     padding-top: 6px;
 }
 
+row check {
+    padding: 1px;
+    border-radius: 12px;
+}
```
danrabbit, line 54: I'd prefer not to do this since we use the rounded shape elsewhere to indicate radio buttons

marbetschar: Personally I think it looks much nicer, especially together with the potential spinner of #104 - but don't have too strong feelings about this. Removed the css style.

src/Widgets/TaskRow.vala
```diff
@@ -75,7 +75,7 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
 
         check = new Gtk.CheckButton ();
         check.valign = Gtk.Align.CENTER;
-        Tasks.Application.set_task_color (source, check);
+        check.get_style_context ().add_provider (taskrow_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
```
danrabbit, line 78: Is this necessary now?

marbetschar: @danrabbit oh boy, seems like it was already late... Good catch! Of course this is no longer needed. Removed.

PR #104, opened 2020-04-08
Showing spinner during task add or update

### LINKED ISSUE #71: Tasks cannot be unchecked
Once checked, a task cannot be unchecked. It will revert to being checked after a couple seconds

marbetschar:
This PR attempts to fix #71 by showing a Spinner while creating/updating a task is active to avoid confusion:

Check and Uncheck

### DIFF
--- a/src/Widgets/TaskRow.vala
+++ b/src/Widgets/TaskRow.vala
@@ -34,6 +34,7 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
     private Gtk.Stack state_stack;
     private Gtk.Image icon;
     private Gtk.CheckButton check;
+    private Gtk.Spinner spinner;
     private Gtk.Entry summary_entry;
     private Gtk.Label description_label;
     private Gtk.Label due_label;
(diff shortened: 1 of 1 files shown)


danrabbit:
@cassidyjames thoughts on this? Do we need to style this so that the spinner appears to be inside a checkbox?

cassidyjames:
Hmm, how long do we expect it to take to process? I feel like showing a spinner at all is probably not necessary unless it's going to take a substantial amount of time; otherwise, we should optimistically show the new state and revert it if it fails for some reason.

marbetschar:
@cassidyjames on my machine it feels like the process takes something about 1-2 seconds. Its not long, but long enough to make the UI feel stuck/slow and the spinner mitigates this feeling a bit.

Personally I prefer the optimistical approach, but for this to work well we need to completely decouple the UI from the EDS backend in a way, which allows the UI to operate on its own. Because unfortunately updating the UI state through EDS callbacks for insert/update/delete a task is rather slow.

FWIW: First steps towards a more independent UI are in the works in PR #100.

marbetschar:
PR #123 provides proper events so we can optimistically show the new state in the UI and revert it, if EDS fails for some reason. That said, this PR is superseeded by #123.

PR #106, opened 2020-04-10
Complete occurence only

### LINKED ISSUE #105: Complete a repeating task
## Prerequisites

## Describe the bug
When we check a recurring task as complete, it does mark the whole series of task as completed which causes the task to "disappear".

### Expected behavior
The task app should just set this one occurrence of the task to "completed". If the user truly wishes to get rid of a recurring task it should be "deleted" instead of "completed". This way, we can handle both scenarios without the need for more features in the UI.

marbetschar:
Avoid having all completed. This Pull Request attempts to fix #105

### DIFF
--- a/src/ListView.vala
+++ b/src/ListView.vala
@@ -159,10 +159,13 @@ public class Tasks.ListView : Gtk.Grid {
     private void on_tasks_added (Gee.Collection<ECal.Component> tasks) {
         tasks.foreach ((task) => {
             var task_row = new Tasks.TaskRow.for_component (task, source);
-            task_row.task_save.connect ((task) => {
-                Tasks.Application.model.update_task (source, task, ECal.ObjModType.ALL);
+            task_row.task_completed.connect ((task) => {
+                Tasks.Application.model.complete_task (source, task, ECal.ObjModType.THIS_AND_PRIOR);
             });
-            task_row.task_delete.connect ((task) => {
+            task_row.task_changed.connect ((task) => {
+                Tasks.Application.model.update_task (source, task, ECal.ObjModType.THIS_AND_FUTURE);
+            });
+            task_row.task_removed.connect ((task) => {
                 Tasks.Application.model.remove_task (source, task, ECal.ObjModType.ALL);
             });
             task_list.add (task_row);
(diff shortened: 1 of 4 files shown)


marbetschar:
Finally got a working solution: The app calculates the next occurence und generates a new task instance for it. It also updates the due date as well as any potential alarm set:

Calculate Next Occurence

alainm23:
Do you think it is necessary to create a new instance of the task?  why not just update the date and send a notificación.

marbetschar:
@alainm23 from what I can tell creating a new "task entity" is how it this situation is expected to be dealt with. At least that's the behaviour one can see working with Apple Reminders and there are suggestions of the exact same behaviour for the Nextcloud web interface:

- https://github.com/nextcloud/tasks/issues/34#issuecomment-353146813
- https://github.com/nextcloud/tasks/issues/34#issuecomment-533230414
- https://github.com/nextcloud/tasks/issues/34#issuecomment-533476436

However, I'm not completely sure if the current implementation is correct. I did a quick comparison between completing a scheduled tasks with this code vs. completing a scheduled task on iOS. As you can see on the following screenshot they look the same in elementary Tasks - but differently on the Nextcloud web interface. Any suggestions on how to make it behave the same way would be appreciated.

Screenshot from 2020-04-11 07-55-31

marbetschar:
I noticed iOS does not only set the completed state, but it also sets the complete percentage to 100 as well as the completed date. I added the code to do the same, but it throws the following error - I guess there is some error in the VAPI. Unfortunately I don't know how to fix this.

```
/usr/include/evolution-data-server/libecal/e-cal-component.h:409:7: note: expected ‘struct icaltimetype *’ but argument is of type ‘struct icaltimetype’
```

The code in question is this one:

https://github.com/elementary/tasks/blob/90feb9ab17239da4a9376ab9a004013125b7936e/src/ListView.vala#L336-L342

marbetschar:
If a task is completed in iOS the following properties are set: status, completed time and complete percentage.

Because elementary Tasks did only set the status to completed, the tasks were displayed differently in the Nextcloud backend.

This has been fixed and recurring tasks which are completed in elementary Tasks now behave the same way as they do in Apple Reminders:

Screenshot from 2020-04-11 17-15-22

PR #120, opened 2020-05-21
Scheduled List

### LINKED ISSUE #46: Add "planned" list

marbetschar:
Replaces #98, Fixes #46.

Contains only the changes necessary for Schedule List based upon latest master and the feedback provided in #98 by @danrabbit.

This PR should be way easier to review.

Screenshot from 2020-05-21 19-29-06

### DIFF
--- a/src/ListView.vala
+++ b/src/ListView.vala
@@ -21,11 +21,49 @@
 public class Tasks.ListView : Gtk.Grid {
     public E.Source? source { get; set; }
 
-    private ECal.ClientView view;
+    private Gee.Collection<ECal.ClientView> views;
+
+    /*
+     * We need to pass a valid S-expression as query to guarantee the callback events are fired.
+     *
+     * See `e-cal-backend-sexp.c` of evolution-data-server for available S-expressions:
+     * https://gitlab.gnome.org/GNOME/evolution-data-server/-/blob/master/src/calendar/libedata-cal/e-cal-backend-sexp.c
+     */
+
+    public void add_view (E.Source task_list, string query) {
+        try {
+            var view = Tasks.Application.model.create_task_list_view (
+                task_list,
+                query,
+                on_tasks_added,
+                on_tasks_modified,
+                on_tasks_removed );
+
+            lock (views) {
+                views.add (view);
+            }
+
+        } catch (Error e) {
+            critical (e.message);
+        }
+    }
+
+    private void remove_views () {
+        lock (views) {
+            foreach (ECal.ClientView view in views) {
+                Tasks.Application.model.destroy_task_list_view (view);
+            }
+            views.clear ();
+        }
+    }
+
+    private Gtk.Revealer settings_button_revealer;
     private EditableLabel editable_title;
     private Gtk.ListBox task_list;
 
     construct {
+        views = new Gee.ArrayList<ECal.ClientView> ((Gee.EqualDataFunc<ECal.ClientView>?) direct_equal);
+
         editable_title = new EditableLabel ();
         editable_title.margin_start = 24;
 
@@ -44,6 +82,10 @@ public class Tasks.ListView : Gtk.Grid {
         settings_button.get_style_context ().add_class (Gtk.STYLE_CLASS_FLAT);
         settings_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DIM_LABEL);
 
+        settings_button_revealer = new Gtk.Revealer ();
+        settings_button_revealer.transition_type = Gtk.RevealerTransitionType.CROSSFADE;
+        settings_button_revealer.add (settings_button);
+
         var placeholder = new Gtk.Label (_("No Tasks"));
         placeholder.show ();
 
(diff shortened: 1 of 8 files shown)


src/ListView.vala
```diff
@@ -144,21 +186,73 @@ public class Tasks.ListView : Gtk.Grid {
...
+        header_label.ellipsize = Pango.EllipsizeMode.MIDDLE;
+        header_label.margin_start = 6;
+
+        row.set_header (header_label);
+    }
+
+    private void on_tasks_added (Gee.Collection<ECal.Component> tasks, E.Source source) {
         tasks.foreach ((task) => {
-            var task_row = new Tasks.TaskRow.for_component (task, source);
+            var task_row = new Tasks.TaskRow.for_component (task, source, "@colorAccent", this.source == null);
```
danrabbit, line 255: Setting the color this way is super overkill, and in the case of the check button it actually does nothing since it is already using accent color by default

src/Widgets/TaskRow.vala
```diff
@@ -334,7 +336,11 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
     }
 
     public void update_request () {
-        Tasks.Application.set_task_color (source, check);
+        if (is_scheduled_view && tint_color != null) {
+            Tasks.Application.set_task_color_from_string (tint_color, check);
```
danrabbit, line 340: This doesn't actually do anything since the only place this is called sets the check to the accent color (which it already is by default)

src/ListView.vala
```diff
@@ -120,14 +144,32 @@ public class Tasks.ListView : Gtk.Grid {
...
+            editable_title.sensitive = false;
+            editable_title.text = _("Scheduled");
+            settings_button_revealer.reveal_child = false;
 
-        task_list.@foreach ((row) => {
-            if (row is Tasks.TaskRow) {
-                (row as Tasks.TaskRow).update_request ();
-            }
-        });
+            Tasks.Application.set_task_color_from_string ("@colorAccent", editable_title);
```
danrabbit, line 152: editable_title already has `Granite.STYLE_CLASS_ACCENT`, so this does nothing

src/Widgets/ScheduledRow.vala
```diff
@@ -0,0 +1,63 @@
...
+        spinner.active = true;
+        spinner.tooltip_text = _("Connecting…");
+
+        status_stack = new Gtk.Stack ();
+        status_stack.add_named (status_image, "image");
+        status_stack.add_named (spinner, "spinner");
+
+        var grid = new Gtk.Grid ();
+        grid.column_spacing = 3;
+        grid.margin_start = 4;
```
danrabbit, line 51: To get this lined up with the SourceRow it looks like you probably want
```suggestion
        grid.margin_start = 10;
```

src/Widgets/ScheduledRow.vala
```diff
@@ -0,0 +1,63 @@
...
+        status_image = new Gtk.Image ();
+        status_image.pixel_size = 16;
+
+        var spinner = new Gtk.Spinner ();
+        spinner.active = true;
+        spinner.tooltip_text = _("Connecting…");
+
+        status_stack = new Gtk.Stack ();
+        status_stack.add_named (status_image, "image");
+        status_stack.add_named (spinner, "spinner");
```
danrabbit, line 47: It doesn't appear that any of this is ever used, so probably shouldn't be here

src/Widgets/ScheduledRow.vala
```diff
@@ -0,0 +1,63 @@
...
+*
+*/
+
+public class Tasks.ScheduledRow : Gtk.ListBoxRow {
+
+    private Gtk.Image icon;
+    private Gtk.Image status_image;
+    private Gtk.Label display_name_label;
+    private Gtk.Stack status_stack;
+    private Gtk.Revealer revealer;
```
danrabbit, line 27: These probably don't all need to be declared ahead of time since they're only used in the construct block

marbetschar: Fixed.

src/Widgets/ScheduledRow.vala
marbetschar: Fixed.

src/Widgets/ScheduledRow.vala
marbetschar: Good catch!

src/ListView.vala
marbetschar: Dropped tint_color property in TaskRow.

marbetschar:
@danrabbit thanks for your feedback, good points! I reworked the code a bit and also added a `Gtk.Stack` to switch between the newly created `scheduled_title` and `editable_title`. The scheduled list title now uses the default accent color which this also improves the contrast level.

PR #125, opened 2020-06-30
Invalidate headers and sorting

### LINKED ISSUE #124: Sorting: Multiple list header for the same grouping
## Describe the bug

The sorting of the list headers does not seem to work properly (see Screenshot below). My guess is this is related to event handling: For each list item, an event is emitted and the corresponding list header is added if there isn't any yet. If two events are emitted for the same day at nearly the same time it might happen we end up with two list headers for the same day:

Screenshot from 2020-06-30 14-05-11

marbetschar:
Forcing resorting and header rendering by calling the corresponding invalidate functions seem to fix #124. Don't know if there is a more efficient approach though.


src/ListView.vala
```diff
@@ -309,7 +309,14 @@ public class Tasks.ListView : Gtk.Grid {
             task_list.add (task_row);
             return true;
         });
-        task_list.show_all ();
+
+        Idle.add (() => {
+            task_list.invalidate_sort ();
+            task_list.invalidate_headers ();
```
danrabbit, line 315: Turns out calling `invalidate_sort` already calls `invalidate_headers`, so you don't need to do both :)

danrabbit (changes requested):

src/ListView.vala
marbetschar: Good catch!

danrabbit (approved):

PR #126, opened 2020-07-31
Port to use LibHandy

lainsce:
This ports Tasks to use LibHandy, and consequently makes it look not broken.

Looks like:
tasks

### DIFF
--- a/data/HeaderBar.css
+++ b/data/HeaderBar.css
@@ -19,28 +19,12 @@
 */
 
 .sidebar-header {
-    background-color: shade (@bg_color, 0.92);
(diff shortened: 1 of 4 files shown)


lainsce:
There's going to be a follow-up PR where I fix header coloring (Scheduled should be using another color, so does the others than default Blue)

alainm23:
What happened to Gtk.Paned?

lainsce:
Gtk.Paned doesn't really work for the full sidebar look and is prone to visual bugs such as its divider breaking the app visually when dragged on the headerbar paned. This is fixed in this PR.

lainsce:
Superceded by #128

PR #128, opened 2020-08-05
MainWindow: Use Hdy.Window and Hdy.HeaderBar

### LINKED ISSUE #121: Use Hdy.ApplicationWindow
This should make the design much simpler where we can have only 1 Gtk.Paned and set the window handles wherever we want them to be. It should also enable us to drop quite a bit of custom css

danrabbit:
supersedes #126 
Fixes #121


marbetschar (approved):
Works as intended. Just a note for any future visitor: By merging this, we will only be able to build Tasks on Odin - since libhandy is not available for Hera.

PR #130, opened 2020-09-15
Dropped support for libecal-1.2

marbetschar:
Since Tasks will arrive on Odin only, I guess there is no need to support libecal-1.2 any longer?


danrabbit (approved):
:+1:

PR #123, opened 2020-06-17
Calendar.Store: Unifying API for Events and Tasks

marbetschar:
This PR is a follow up of the work done in wingpanel, to use one service object providing access to events and tasks using a unifyied API:

https://github.com/elementary/wingpanel-indicator-datetime/pull/225

PR #131, opened 2020-09-16
Use same code structure as in Calendar

marbetschar:
This PR does it make easier for future contributors to provide fixes for Tasks as well as Calendar, because we are now using the same code structure. Also this PR provides the base for later wingpanel-indicator-datetime integration of Tasks.

Changes in a Nutshell:

- Renamed `Tasks.TaskModel` to `Calendar.TaskStore`
- Moved EDS specific utility functions to `Calendar.Util`
- Made sure the EDS specific utility functions code is equal to the ones used in the Calendar app


marbetschar:
btw: as concluded [here](https://github.com/elementary/calendar/pull/588#issuecomment-692160733). there is no need to further merge the code of Calendar.EventStore and Calendar.TaskStore for now. Therefore this PR superseds https://github.com/elementary/tasks/pull/123 as well.

meson.build
```diff
@@ -20,13 +20,16 @@ executable(
     'src/Application.vala',
     'src/ListView.vala',
     'src/MainWindow.vala',
+    'src/Services/Calendar/TaskStore.vala',
+    'src/Services/Calendar/Util/DateTime.vala',
+    'src/Services/Calendar/Util/ECalComponent.vala',
+    'src/Services/Calendar/Util/ICalTime.vala',
```
danrabbit, line 26: This folder structure seems a little excessively nested

There's also currently an existing "Util.vala". All the Util classes should probably be in the same folder and I guess this needs to be renamed to be clear how it is different from these other Util classes

src/Services/Calendar/TaskStore.vala
```diff
@@ -19,12 +19,12 @@
...
 
 
-errordomain Tasks.TaskModelError {
+errordomain Calendar.TaskStoreError {
     CLIENT_NOT_AVAILABLE
 }
 
 
-public class Tasks.TaskModel : Object {
+public class Calendar.TaskStore : Object {
```
danrabbit, line 27: Does it make sense to namespace this under Calendar for the Tasks app?

danrabbit (changes requested):
This folder structure seems a little excessively nested and it appears that a few of these classes only contain one method. Is it really necessary to split this into multiple files?

If it is really necessary, all of the util classes should probably be grouped together. There's still the existing "Util.vala" that should be renamed to be clear how it is different from these other Util classes.

I'm not certain how this branch makes Tasks more maintainable. It seems to be creating a lot of complexity

src/Services/Calendar/TaskStore.vala
marbetschar: Ideally the namespace should be `EDS` or similar for the Calendar as well as for the Tasks app. I did not change this yet, to avoid further confusion.

marbetschar:
@danrabbit 

> This folder structure seems a little excessively nested and it appears that a few of these classes only contain one method. Is it really necessary to split this into multiple files?

It's not strictly necessary - but I figured I follow the same code structure as we have it in the Calendar app. It'll be easier to add the code for both (Tasks and Events) to WIngpanel IMHO, if we have a shared namespace `EDS.Util`.

> If it is really necessary, all of the util classes should probably be grouped together. There's still the existing "Util.vala" that should be renamed to be clear how it is different from these other Util classes.

The difference is, the existing `Util.vala` contains Tasks specific utility functions - therefore its namespace: `Tasks.Util`.

> I'm not certain how this branch makes Tasks more maintainable. It seems to be creating a lot of complexity

marbetschar:
@danrabbit the goal of the code refactoring is outlined below.

As said, this is not strictly necessary - but my feeling is it makes maintenance of Tasks, Calendar and Wingpanel easier, especially once we also integrated TaskStore into Wingpanel. Of course we can also combine all `EDS/Util/*` files into one `EDS/Util.vala` file, if thats preferrred (personally I'd always go for more files with less content, cause IMHO its easier to understand - but that's just my personal taste).

Also, feel free to close this PR if you don't see fit. No hard feelings here.

mermaid-diagram-20200930132248

marbetschar:
FWIW: We can also strip the whole `EDS.Util` namespace and put everything into `EDS` - maybe thats even better to avoid confusion with `Tasks.Util` and `Calendar.Util` respectively...?

marbetschar:
@danrabbit how about something like this?

danrabbit:
Hey sorry it took a while to get back to this!

I feel like namespacing it `EDS` is confusing here because it seems like this is a library namespace when it's actually part of this code base. I can imagine a future contributor confused searching Valadoc for this namespace and wondering what is going on. I think historically, we've using `Common` here, but again historically this doesn't end up synced across repos and it turns out to be a lot of added complexity for no actual gain

I think as we talked in our standup a few weeks ago that if things are really commonly used by multiple repos then we should pursue upstreaming those utilities into Evolution. That would really be the best way to maintain this common code.

marbetschar:
Fair point!

Sorry, for keeping you busy with this - still trying to wrap my head around how things fit together in terms of how the ecosystem and community works.

Closing this PR.

PR #119, opened 2020-05-15
Add missing palette colors to color button list

4jNsY6fCVqZv:
Fixes #117

@cassidyjames Can you tell me where I can find the hexadecimal value of the checkmark white? Then I'll look at the contrasts, too.

PR #137, opened 2020-09-21
ListSettingsPopover: Add Mint and Bubblegum

danrabbit:
Supersedes #119


4jNsY6fCVqZv:
Is there a reason why you write "Add Mint and Bubblegum" but then introduce a new color name with pink?

danrabbit:
@4jNsY6fCVqZv these aren't user-facing strings and you are missing the context of seeing the color, so it's clearer for a developer to keep it in line with the name of the style class provided by the system stylesheet

4jNsY6fCVqZv:
@danrabbit Then why is there any distinction at all? Wouldn't it be easier to always use the names of the style class? Or to adapt the style classes to the names of the color palette? Then it would be coherently solved at least in my perception.

marbetschar (approved):
Looks good to me!

PR #142, opened 2020-10-01
po: Rename io.elementary.reminders.pot into io.elementary.tasks.pot

tintou:


ryonakano (approved):
I guess the content in the file should be updated automatically by the gettext actions.

PR #143, opened 2020-10-05
Fixes for ListView Design

marbetschar:
This PR fixes some styling issues which arised on Odin. What I also like to fix is the background color of the list content - but I did not figure out how to do it (see black arrow in 2nd screenshot).

**Before**

Screenshot from 2020-10-05 19-25-09

**After**

Screenshot from 2020-10-05 20-03-03

### DIFF
--- a/data/TaskRow.css
+++ b/data/TaskRow.css
@@ -32,17 +32,12 @@ row {
     transition: all 250ms ease-in-out;
 }
 
-row:focus:not(.card) {
-    background-color: shade (@base_color, 0.8);
-}
-
 .past-due .due-date {
     background: alpha (@error_color, 0.2);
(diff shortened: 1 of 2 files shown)


danrabbit:
I think these are probably bugs in Greenfield that should be addressed there.

At least the background thing is fixed by https://github.com/elementary/stylesheet/pull/844

marbetschar:
You were right - the background style is indeed an issue in Greenfield and your change at https://github.com/elementary/stylesheet/pull/844 does fix it. For the flat entry further adjustments are needed as proposed in this PR (see screenshots):

**Before**

Screenshot from 2020-10-06 08-32-01


**After**

Screenshot from 2020-10-06 08-35-36

PR #145, opened 2020-10-06
MainWindow: Connect select_row before selection

danrabbit:
This fixes restoring the last selected task list. We must connect to the select_row before making the selection. Otherwise the first selection is missed


marbetschar (approved):
Nice catch! For whatever reason this does not yet seem to work for the Scheduled list - but there I guess is something else going wrong.

PR #147, opened 2020-10-07
Fixes missing scheduled list tasks

marbetschar:
Fixes an issue, where not all scheduled tasks are shown in the Scheduled List. The root cause of the issue is the s-expression for the Scheduled List in `MainWindow.vala`:

```
144:    var query = "AND (NOT is-completed?) (OR (has-start?) (has-alarms?))";
```

Unfortunately, we can't query for ''(has-due?)" [because EDS does not support it](https://gitlab.gnome.org/GNOME/evolution-data-server/-/blob/master/src/calendar/libedata-cal/e-cal-backend-sexp.c#L1242-1265).

So we basically end up with the following two choices for the s-expression:

1. `(has-start?)`
2. `(has-alarms?)`

Unfortunately, `(has-alarms?)` does not qualify. Because "Location-Based-Reminder"'s on iOS (and on other devices) essentially create an alarm with location information - so location based Tasks would end up in the Scheduled List too. This is not what we want.

**Long story short:** The only way to retrieve all scheduled tasks is to query for `(has-start?)` and make sure, the Tasks app sets the start date to the same date used for the due date.

That's essentially what I ended up doing in this PR.


marbetschar:
FWIW: Asked for `has-due?` support [@upstream](https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/254).

danrabbit:
Okay, so this is actually interesting. Things I've added from Tasks with a schedule do not show up in the Scheduled view. But, anything I've added from Reminders on iOS does. So maybe this is actually an issue with how we're saving tasks and not how we're reading them

marbetschar:
@danrabbit yes, tasks need dtstart set in order to show up in the Scheduled List. Thats exactly what iOS Reminders does and what this PR fixes.

danrabbit (approved):
Ahh I misunderstood. Yup this does the thing

PR #150, opened 2020-10-09
CalDAV: Add Task List for WebDAV backend

### LINKED ISSUE #148: Use selected section for new list's parent
## Feature

Thanks to #13 we are now able to add new task lists. Although, all of them are added locally as of now. Therefore I'm wondering if we should grab the information for the new `E.Source`'s parent from the "selected section":

192-67-max

In the scenario above, we can argue that the user "selected" the section "Development" because the list "Any List" is selected underneath it. Therefore, `<Ctrl> + N` should create a new list in "Development".

### Edge Cases

If we implement this, we need to account for the situation where the user selected the "Scheduled" list - or a section does not contain a task list at all:

**If "Scheduled" list is selected**

In this case `<Ctrl> + N` should just open the popover described in the next case.

**If a section is not "selectable" (does not contain any task lists)**

In this case, the only option the user has is to click on "+ Add Task List...". There a popover shows all sections and the user is able to select where to add the new list. Similar to what we have on iOS:

518-269-max

a solution in case the user selected the "Scheduled" list -

marbetschar:
This PR fixes #148.


marbetschar:
although I'm able to add a new list using the change in this PR, I'm not able to add tasks to the newly created list (getting permission denied errors). Also the list does not appear in the Nextcloud backend.

@danrabbit how does this PR behave on your end? You are using a Google backend, right?

danrabbit:
Yea for GTasks I'm getting a freeze. Opening it in Evolution, it looks like there's no username associated with the list and it doesn't like that. 

After adding a username, it seems to be working as expected, but it appears to not be able to be deleted which is interesting

marbetschar:
The approach always throws a  "Data source '...' does not support creating remote resources" - for both, Nextcloud and GTasks. Its not possible to add a new task list from within Evolution as well (at least I'm not able to figure out how).

@tintou any chance you can point me into the right direction? Or is there simply no way for creating a new E.Source using the these backends?

marbetschar:
As of now it is not possible to create a new task list for a WebDAV backend using libecal. If we want to integrate this functionality right now, we need to use lower level API's.

See: https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/261

marbetschar:
I guess GTasks does not yet work (not tested). But this does the trick for WebDAV backends (e.g. Nextcloud). So I suggest we get this merged and create a new issue for GTasks. What do you think @danrabbit?

danrabbit:
I feel like this method for choosing the list to add to might feel a little non-deterministic and as far as I can tell makes it impossible to create local lists unless you have one already. It would probably be better to show a popover with a list of all of the connected sources so the user can choose which one they want to add to explicitly

marbetschar:
@danrabbit I agree and that was exactly the plan. But I do think we should use the proposed method here for the `Ctrl + N` shortcut - otherwise the user always has to click twice to add a new list.

marbetschar:
also, the `Ctrl + N` shortcut can fall back to open the popover if it can't determine what to do (i.e. there is no list or the scheduled list is selected)

marbetschar:
the current approach of figuring out the WebDAV backend url does not compile with the following error. In the header file there is an additional `E.Source` argument declared, and this seems to be needed:

```
../src/MainWindow.vala:281:3: error: too few arguments to function ‘e_webdav_discover_sources_finish’
  281 |                                     E.webdav_discover_sources.end (
      |   ^                               
In file included from /usr/include/evolution-data-server/libedataserver/libedataserver.h:100,
                 from io.elementary.tasks@exe/src/MainWindow.c:29:
/usr/include/evolution-data-server/libedataserver/e-webdav-discover.h:61:10: note: declared here
   61 | gboolean e_webdav_discover_sources_finish (ESource *source,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

But if I add the missing E.Source, the code no longer compiles with the following error:

```
../src/MainWindow.vala:281.37-288.37: error: 1 extra arguments for `bool E.webdav_discover_sources.end (GLib.AsyncResult, out string, out GLib.TlsCertificateFlags, out GLib.SList<E.WebDAVDiscoveredSource?>, out GLib.SList<string>)
```

marbetschar:
Now the binding seems to be fixed - but the app crashes with a `free(): invalid pointer` error. FWIW: This error occurs for the `webdav_discover_sources_sync` method as well. Here's the gdb backtrace:

```
...
[Detaching after vfork from child process 169710]
[New Thread 0x7fffd4be2700 (LWP 169716)]
free(): invalid pointer

Thread 1 "io.elementary.t" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) backtrace
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007ffff6d75859 in __GI_abort () at abort.c:79
#2  0x00007ffff6de03ee in __libc_message (action=action@entry=do_abort, fmt=fmt@entry=0x7ffff6f0a285 "%s\n") at ../sysdeps/posix/libc_fatal.c:155
#3  0x00007ffff6de847c in malloc_printerr (str=str@entry=0x7ffff6f084ae "free(): invalid pointer") at malloc.c:5347
#4  0x00007ffff6de9cac in _int_free (av=<optimized out>, p=<optimized out>, have_lock=0) at malloc.c:4173
#5  0x0000555555567385 in _g_free0_ (var=0x555555978430) at ../src/MainWindow.vala:281
#6  0x00007ffff7ef0ab0 in g_slist_foreach () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#7  0x00007ffff7ef0adf in g_slist_free_full () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#8  0x00005555555673af in _g_slist_free__g_free0_ (self=0x555555a45e50 = {...}) at ../src/MainWindow.vala:281
#9  0x00005555555681fd in ____lambda62_ (_data6_=0x55555577b060, obj=0x7fffd80ac6d0, res=0x555555f31600) at ../src/MainWindow.vala:274
#10 0x0000555555568285 in _____lambda62__gasync_ready_callback (source_object=0x7fffd80ac6d0, res=0x555555f31600, self=0x55555577b060)
    at ../src/MainWindow.vala:268
#11 0x00007ffff72d4f49 in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#12 0x00007ffff72d4f8d in  () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#13 0x00007ffff7ed0e8e in g_main_context_dispatch () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#14 0x00007ffff7ed1240 in  () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#15 0x00007ffff7ed12e3 in g_main_context_iteration () at /lib/x86_64-linux-gnu/libglib-2.0.so.0
#16 0x00007ffff7302fd5 in g_application_run () at /lib/x86_64-linux-gnu/libgio-2.0.so.0
#17 0x0000555555562694 in tasks_application_main (args=0x7fffffffddb8, args_length1=1) at ../src/Application.vala:113
#18 0x00005555555626db in main (argc=1, argv=0x7fffffffddb8) at ../src/Application.vala:111
```

meson.build
```diff
@@ -12,6 +12,25 @@ gresource = gnome.compile_resources(
...
+    dependency('gobject-2.0'),
+    dependency('granite', version: '>=0.5'),
+    dependency('gtk+-3.0'),
+    libecal_dep,
+    dependency('libedataserver-1.2'),
+    dependency('libhandy-1', version: '>=0.90.0'),
+    dependency('libical-glib')
+]
+
+if libecal_dep.version().version_compare('<=3.38.1')
```
tintou, line 30:
```suggestion
if libecal_dep.version().version_compare('<=3.39')
```
there won't be backports of the these until next EDS release

marbetschar:
@elementary/desktop-developers a review of anyone using a CalDAV backend to synchronize tasks would be great :+1: 

This PR contains the following changes:

- added a popover to the "Add Task List" button, which provides a selection of the available collections
- removed `Ctrl+n` accelerator, since it made no longer sense (the user needs to define in which collection the task list should be added)
- Made it possible to add a task list via WebDAV (this requiried some VAPI workarounds - the needed fixes are already upstreamed)
- Added some low-level plumbing to distinguish the backend which should be used

src/MainWindow.vala
```diff
@@ -370,6 +356,83 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+        lock (collection_sources) {
+            foreach (var collection_source in collection_sources) {
+                var source_button = new Gtk.ModelButton () {
+                    text = get_collection_source_display_name (collection_source)
+                };
+
+                source_button.button_release_event.connect (() => {
+                    add_new_list (collection_source);
+                    return Gdk.EVENT_PROPAGATE;
+                });
```
danrabbit, line 410: If we're not worried about stopping the popover from being closed we can do:
```suggestion
                source_button.clicked.connect (() => {
                    add_new_list (collection_source);
                });
```

src/MainWindow.vala
```diff
@@ -370,6 +356,83 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+            }
+            collection_sources.add (collection_source);
+        }
+        update_add_tasklist_buttonbox ();
+    }
+
+    private void update_add_tasklist_buttonbox () {
+        foreach (var child in add_tasklist_buttonbox.get_children ()) {
+            add_tasklist_buttonbox.remove (child);
+        }
```
danrabbit, line 399: This seems really inefficient to remove all the items and then add them back. It would probably be better to have separate add and remove functions if you're worried about making sure this list doesn't contain things that have been removed

src/MainWindow.vala
```diff
@@ -370,6 +356,83 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+
+            if (collection_sources.contains (collection_source) || !collection_source.enabled || !collection_source_tasklist_extension.selected) {
+                return;
+            }
+            collection_sources.add (collection_source);
+        }
+        update_add_tasklist_buttonbox ();
+    }
+
+    private void update_add_tasklist_buttonbox () {
```
danrabbit, line 396: This seems to only ever be called from `add_collection_of_source ()`, so I guess this probably doesn't need to be its own function

src/MainWindow.vala
```diff
@@ -95,14 +95,15 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
         scrolledwindow.hscrollbar_policy = Gtk.PolicyType.NEVER;
         scrolledwindow.add (listbox);
 
-        var add_tasklist_button = new Gtk.Button () {
-            action_name = ACTION_PREFIX + ACTION_ADD_NEW_LIST,
-            always_show_image = true,
+        add_tasklist_popover = new Gtk.Popover (null);
+        add_tasklist_buttonbox = new Gtk.ButtonBox (Gtk.Orientation.VERTICAL);
+        add_tasklist_popover.add (add_tasklist_buttonbox);
```
danrabbit, line 100: Can we keep like things grouped please:
```suggestion
        add_tasklist_buttonbox = new Gtk.ButtonBox (Gtk.Orientation.VERTICAL);

        add_tasklist_popover = new Gtk.Popover (null);
        add_tasklist_popover.add (add_tasklist_buttonbox);
```

src/MainWindow.vala
```diff
@@ -33,7 +31,10 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
     private uint configure_id;
     private Gtk.ListBox listbox;
     private Gee.HashMap<E.Source, Tasks.SourceRow>? source_rows;
+    private Gee.Collection<E.Source>? collection_sources;
     private Tasks.ListView listview;
+    private Gtk.Popover add_tasklist_popover;
```
danrabbit, line 36: Does this need to be scoped to the whole class?

vapi/libecal-2.0-fixes.vapi
```diff
@@ -1,4 +1,9 @@
 namespace E {
+	[CCode (cheader_filename = "libedataserver/libedataserver.h", cname = "e_webdav_discover_free_discovered_sources")]
+	[Version (since = "3.18")]
+	public static void webdav_discover_do_free_discovered_sources (owned GLib.SList<E.WebDAVDiscoveredSource?> discovered_sources);
+    [CCode (cname = "e_webdav_discover_sources_finish")]
+    bool webdav_discover_sources_finish (E.Source source, GLib.AsyncResult result, out string out_certificate_pem, out GLib.TlsCertificateFlags out_certificate_errors, out GLib.SList<E.WebDAVDiscoveredSource?> out_discovered_sources, out GLib.SList<string> out_calendar_user_addresses) throws GLib.Error;
```
danrabbit, line 6: Looks like we have some indentation inconsistency here

danrabbit (changes requested):
This appears to still work for local task lists, but it does not appear to work for Google tasks and my iCloud/CalDAV source doesn't show at all in the list.

We should probably make sure we list all sources and then mark unsupported backends as insensitive. This way it doesn't look like a bug that they're missing and we're proactive about showing that they can't be added to instead of letting the user choose and get an error dialog

I left some other comments/questions inline

marbetschar:
@danrabbit indeed, this PR does not yet work for GTasks. There is a [dedicated issue for GTask](https://github.com/elementary/tasks/issues/164). However, proactively showing all backends and make unsupported ones insensitive is a great idea!

Regarding iCloud/CalDAV: I'm not sure why this does not show up. Need to investigate...

marbetschar:
I made the backend discovery simpler by just looping over `registry.list_sources (E.SOURCE_EXTENSION_COLLECTION);` once at application start. Because those rarely change, I guess this is good enough for now. Also, this *should* include everything - that said iCloud should appear too now. But I was not able to verify this yet.

For whatever reason, the local backend is not present in the mentioned `list_sources` call, so I had to address this separately with `add_collection_source (registry.ref_builtin_task_list ());`.

danrabbit:
So now with this branch, the only item I have is an insensitive item for GTasks, no local or iCloud xD

marbetschar:
@danrabbit I just tested with my iCloud account. I added it as "Collection Account" in Evolution, using the caldav.icloud.com Server and a newly created Application Password as you described [here](https://github.com/elementary/switchboard-plug-onlineaccounts/issues/7#issuecomment-519142871).

The current WebDAV implementation works for iCloud as well as for Nextcloud on my end. I'm able to create new task lists as well as renaming them:

iCloud

marbetschar:
@danrabbit after I double checked by logging into icloud.com via Webbrowser, I'm not able to see the before created task lists. After a bit of research, it seems like Apple shifted gears on iOS 13 how Reminders are synced with iCloud.com because of new features. [This caused quite some trouble](https://github.com/nextcloud/server/issues/17190).

So there might be two different scenarios to deal with: Already upgraded Reminders, and not yet upgraded Reminders: https://support.apple.com/en-us/HT210220

IMHO its a bit nebulous what the implications are. Maybe you're more familiar with this?

marbetschar:
I created a [dedicated issue for iCloud.com support](https://github.com/elementary/tasks/issues/178), so we can discuss the details over there. From the WebDAV perspective I'd claim this PR works. Therefore we can focus on Code review in this PR.

danrabbit (approved):
Works as expected for Local lists. Sounds like GTasks and iCloud are gonna be separate issues.

Nice work :)

PR #153, opened 2020-10-10
Rename Task List

marbetschar:
The new task list name does not get stored in the Nextcloud / CalDAV backend. This PR fixes this.


src/ListView.vala
```diff
@@ -192,7 +192,24 @@ public class Tasks.ListView : Gtk.Grid {
...
             source.display_name = editable_title.text;
-            source.write.begin (null);
+
+            if (source.writable) {
+                // if (source.has_extension (E.SOURCE_EXTENSION_WEBDAV_BACKEND)) {
+            //     var source_webdav_extension = (E.SourceWebdav) source.get_extension (E.SOURCE_EXTENSION_WEBDAV_BACKEND);
+            //     source_webdav_extension.display_name = editable_title.text;
+            // }
+            //
+                source.write.begin (null, (obj,res) => {
```
alainm23, line 202: Add a space after `obj,`
`source.write.begin (null, (obj, res) => {`

src/ListView.vala
```diff
@@ -191,8 +191,48 @@ public class Tasks.ListView : Gtk.Grid {
...
         editable_title.changed.connect (() => {
-            source.display_name = editable_title.text;
-            source.write.begin (null);
+            Tasks.Application.model.get_registry.begin ((obj, res) => {
+                try {
+                    var registry = Tasks.Application.model.get_registry.end (res);
+                    var collection_source = registry.find_extension (source, E.SOURCE_EXTENSION_COLLECTION);
+
+                    if (collection_source != null && source.has_extension (E.SOURCE_EXTENSION_WEBDAV_BACKEND)) {
+                        debug (@"WebDAV Rename '$(source.get_uid())'");
```
danrabbit, line 200: Can we use printf syntax here please:
```suggestion
                        debug ("WebDAV Rename: %s", source.get_uid ());
```

src/ListView.vala
```diff
@@ -191,8 +191,48 @@ public class Tasks.ListView : Gtk.Grid {
...
+
+                        var collection_source_webdav_session = new E.WebDAVSession (collection_source);
+                        var source_webdav_extension = (E.SourceWebdav) source.get_extension (E.SOURCE_EXTENSION_WEBDAV_BACKEND);
+
+                        var credentials_provider = new E.SourceCredentialsProvider (registry);
+                        E.NamedParameters credentials;
+                        credentials_provider.lookup_sync (collection_source, null, out credentials);
+                        collection_source_webdav_session.credentials = credentials;
```
danrabbit, line 208: We can probably organize this is a bit more to keep like things together
```suggestion
                        var source_webdav_extension = (E.SourceWebdav) source.get_extension (E.SOURCE_EXTENSION_WEBDAV_BACKEND);

                        E.NamedParameters credentials;
                        var credentials_provider = new E.SourceCredentialsProvider (registry);
                        credentials_provider.lookup_sync (collection_source, null, out credentials);

                        var collection_source_webdav_session = new E.WebDAVSession (collection_source);
                        collection_source_webdav_session.credentials = credentials;
```

src/ListView.vala
```diff
@@ -191,8 +191,48 @@ public class Tasks.ListView : Gtk.Grid {
...
+                            collection_source_webdav_session,
+                            source_webdav_extension.soup_uri.to_string (false),
+                            changes,
+                            null
+                        );
+
+                        registry.refresh_backend_sync (collection_source.uid, null);
+
+                    } else {
+                        debug (@"Local Rename '$(source.get_uid())'");
```
danrabbit, line 227:
```suggestion
                        debug ("Local Rename: %s", source.get_uid ());
```

danrabbit (changes requested):

src/ListView.vala
```diff
@@ -191,8 +191,49 @@ public class Tasks.ListView : Gtk.Grid {
...
+                        registry.refresh_backend_sync (collection_source.uid, null);
+
+                    } else {
+                        debug ("Local Rename: %s", source.get_uid ());
+                        source.display_name = editable_title.text;
+                        registry.commit_source_sync (source, null);
+                    }
+
+                } catch (Error e) {
+                    critical (e.message);
```
danrabbit, line 234: If we completely failed to rename the list, I think we should probably throw an error to the user

src/ListView.vala
```diff
@@ -191,8 +191,49 @@ public class Tasks.ListView : Gtk.Grid {
...
+                            collection_source_webdav_session,
+                            source_webdav_extension.soup_uri.to_string (false),
+                            changes,
+                            null
+                        );
+
+                        registry.refresh_backend_sync (collection_source.uid, null);
+
+                    } else {
+                        debug ("Local Rename: %s", source.get_uid ());
```
danrabbit, line 228: I also wonder if we should throw an error here. This means that we can't rename the list on the server right? So if a user renames a list and then goes to another device, they will be looking for a renamed list that doesn't exist. I would imagine we should inform them of that

danrabbit (changes requested):

marbetschar:
@danrabbit can you please re-review? I've done the following changes, including your suggestions:

- Moved the complexity of renaming a task list for different kind of backends into the `TaskModel`
- Throwing an error if we are not able to rename
- Display the error in a `Granite.MessageDialog`

I also had to change the `EditableLabel` a little bit,to avoid get the `changed` signal fired twice if the user hits the Enter key to submit the new task list name (aka "end editing in EditableLabel).

marbetschar:
@danrabbit friendly reminder :)
would be great if you review/merge these changes. If we're good here, I'll adapt PR #150  accordingly.

src/TaskModel.vala
```diff
@@ -147,6 +148,48 @@ public class Tasks.TaskModel : Object {
         }
     }
 
+    public async void update_task_list_display_name (E.Source task_list, string display_name) throws Error {
+        var registry = get_registry_sync ();
+        var collection_source = registry.find_extension (task_list, E.SOURCE_EXTENSION_COLLECTION);
+
+        if (collection_source != null && task_list.has_extension (E.SOURCE_EXTENSION_WEBDAV_BACKEND)) {
+            debug (@"WebDAV Rename '$(task_list.get_uid())'");
```
danrabbit, line 156: printf style please :) 
```suggestion
            debug ("WebDAV Rename", task_list.get_uid());
```

src/TaskModel.vala
```diff
@@ -147,6 +148,48 @@ public class Tasks.TaskModel : Object {
...
+                collection_source_webdav_session,
+                source_webdav_extension.soup_uri.to_string (false),
+                changes,
+                null
+            );
+
+            registry.refresh_backend_sync (collection_source.uid, null);
+
+        } else if (task_list.parent == "local-stub") {
+            debug (@"Local Rename '$(task_list.get_uid())'");
```
danrabbit, line 183: printf style here too

danrabbit (changes requested):
Couple of small code style comments

I can't really test this because it doesn't work for my CalDav or GTasks lists, but at least the error messaging is working :)

marbetschar:
@danrabbit fixed the requested code bits.

Yes, for some reason the WebDAV specific solution does not seem to work for the Google backend. Need to figure out how to do it there and [created an issue](https://github.com/elementary/tasks/issues/163) to keep track of it.

At least for regular WebDAV this solution works - so we should be fine for backends such as Nextcloud.

danrabbit (approved):

Issue #152, opened 2020-10-09
Task View is very big.


alainm23:
## Prerequisites
- [x] I have searched open and closed issues for duplicates.

## Feature
I understand that you use the same code to create and edit a task but the design is very large and takes up about half the space of the application. I currently use a laptop with a 15.6 inch screen and a resolution of 1366 * 768 (16: 9)

**Describe the solution you'd like**
Make the task view smaller. 

**Existing work**
**macOS Reminders:** Use a simple line for task and notes.
![image](https://user-images.githubusercontent.com/33765137/95601384-fa506e00-0a18-11eb-9448-0bef5ebe08a0.png)

**Todoist:** Use a simple line for task name and actions buttons to select a date, reminder, etc.
![image](https://user-images.githubusercontent.com/33765137/95601521-2ec42a00-0a19-11eb-9b72-e780f6c4824b.png)

**Planner** Use a simple line for task name and actions buttons to select a date, reminder, etc.
![image](https://user-images.githubusercontent.com/33765137/95601751-86fb2c00-0a19-11eb-8a86-a99dd502b1a3.png)

**Additional context**
![image](https://user-images.githubusercontent.com/33765137/95601098-a34a9900-0a18-11eb-8a65-fca700aa14f5.png)

danrabbit:
Yeah the biggest offender here is that the comment area is so large

I like Todois'st/Planner's approach of having image buttons to add additional information like comments, schedules, alarms, etc.

alainm23:
Comments?

![image](https://user-images.githubusercontent.com/33765137/97953199-ad518480-1d97-11eb-841c-f29689c7fcad.png)

PR #85, opened 2020-02-13
Some progress toward due date editing

danrabbit:

PR #157, opened 2020-10-24
New EntryPopover supporting Location & DateTime data types

marbetschar:
This PR is an initial attempt to improve the UX of Task editing (https://github.com/elementary/tasks/issues/152). I created a generic EntryPopover. In this PR I show case its usage with the newly created `DateTimePopover` and `LocationPopover` Widgets (screenshots below). Please note the `LocationPopover` is not fully complete yet, it needs some more work. Also the design details can use some more polish - but you get the idea.

Screenshot from 2020-10-24 16-06-26

Screenshot from 2020-10-24 16-06-49

@danrabbit is this related to https://github.com/elementary/tasks/pull/85?


tintou:
That looks really cool

tintou:
I often use the location field to set a URL, I don't know if that's just me or standard (in which case we should support this and don't show a map)

marbetschar:
My intention was to allow arbitrary text in the Location field, so one can also write "Skype" or "Hangouts". Obviously this would also allow URL's. IMHO there are cases where we do want to set both, URL and location - therefore I don't think we should officially claim the location field should be used for URLs.

Isn't there a dedicated URL field in EDS? macOS Calendar does also distinguish these two fields:

Week-New

marbetschar:
Just did a bit of digging, in the standard there are two fields (location and url). So I think we should also reflect this in our Tasks app. But we can still allow arbitrary text in the location field of course.

**VTODO**

```
todoprop   = *(
                  ...
                  ; The following are OPTIONAL,
                  ; but MUST NOT occur more than once.
                  ;
                  ... / location / ... / url / ...
```

Source:
- [RFC5545 3.6.2.  To-Do Component](https://tools.ietf.org/html/rfc5545#section-3.6.2)

marbetschar:
trying to improve the styling of the EntryPopover - but I don't seem to be able to set the font color. In addition, the background-color does not get applied on the whole Widget, seems like if not clicked some other color is applied above. Anyone an idea how to fix this?

Styling for EntryPopover

alainm23:
> trying to improve the styling of the EntryPopover - but I don't seem to be able to set the font color. In addition, the background-color does not get applied on the whole Widget, seems like if not clicked some other color is applied above. Anyone an idea how to fix this?
> 
> Styling for EntryPopover
> 

Try:
`
.button-css label {
    color: red;
}
`

marbetschar:
The LocationPopover now works as well. Functionwise this is ready to merge.
Maybe someone can help with styling? Unfortunately I was not able to make the suggestion of Alain work.

danrabbit:
Can you split this up into two branches? One for the date/time stuff and one for the location stuff? 1,000 line diff is a little bit much to review

jlnr:
> macOS Calendar does also distinguish these two fields...

But that's only in the Calendar app; the Apple Reminders (~ elementary Tasks) has a single field for "Location".

marbetschar:
@jlnr indeed, Apple Reminders allows to set a location to get notified on arrival or departure - but [as it turned out](https://github.com/elementary/tasks/blob/f3817889b4c78e7cfc9f3a380668ef5f2ccb2c05/src/Util.vala#L165-L196), this "notification-location" is **not** stored in the CalDAV "Location" field. The notification-location in Apple specific custom fields (`X-APPLE-PROXIMITY`, `X-APPLE-STRUCTURED-LOCATION`) within a VALARM (aka. "notification") which is in turn embedded in the VTODO (aka. "task").

I'm wondering how the behaviour is like on Android - but it seems there isn't a common ground here. So things will get messy 🥴

marbetschar:
this PR is superseded by #159. Will create a new PR for LocationPopover once we merged the DateTimePopover.

marbetschar:
Superseded by #174 and #159 . Closing.

PR #159, opened 2020-10-29
DateTimePopover Entry

marbetschar:
As requested in #157 this branch only adds the DateTimePopover entry (Location will follow as dedicated PR once this is merged). There are some minor styling issues left, maybe someone can help out?


danrabbit (changes requested):
This is looking really good!

I made a branch with some code style changes and CSS fixes. I figured that would be a lot easier than trying to leave a review for a lot of those things: https://github.com/elementary/tasks/pull/162

Inside the popover, I wonder if it would be better to throw in a separator and then put the "Today button" on its own line and use Gtk.ModelButton instead. At the moment it looks like it somewhere effects the time entry instead of being a standalone shortcut. This would also make sense if we added more options like "Tomorrow" some other shortcut

I don't think we should make the widget insensitive when the card is collapsed. It's kind of nice and convenient to be able to quickly change or remove the schedule without going into the whole full edit mode. So I think we just need to make sure we update the revealer whether or not we're in the edit mode

which leads me to also wonder if we should make the revealer as part of EntryPopover so we don't have to re-implement this for every one we add, since presumably they will all exhibit the same behavior of hiding unless they contain a value

It doesn't look like `direction` is used either here or in the maps branch, so I'm wondering if there's a reason to add this?

That's all for now :) Thanks for working on this!

marbetschar:
thanks for all of your feedback, really appreciate it! 

> Inside the popover, I wonder if it would be better to throw in a separator and then put the "Today button" on its own line and use Gtk.ModelButton instead. At the moment it looks like it somewhere effects the time entry instead of being a standalone shortcut. This would also make sense if we added more options like "Tomorrow" some other shortcut

Valid point. I changed the popover to use a ModelButton (before/after screenshots below).

> I don't think we should make the widget insensitive when the card is collapsed. It's kind of nice and convenient to be able to quickly change or remove the schedule without going into the whole full edit mode. So I think we just need to make sure we update the revealer whether or not we're in the edit mode

The reasoning for insensitivity is to avoid endless `value_changed` signal loops if we set the value programmatically i.e. when the task is initially loaded. Because this happens after the events are bound. Wasn't aware of another way, so I chose to use the `sensitive` flag. Open for better solutions though.

> which leads me to also wonder if we should make the revealer as part of EntryPopover so we don't have to re-implement this for every one we add, since presumably they will all exhibit the same behavior of hiding unless they contain a value

Valid point, thought about that too. Wasn't sure if that's too much then though. Seems to make sense - but I guess we need to make `transition_type` configurable then, right?

> It doesn't look like direction is used either here or in the maps branch, so I'm wondering if there's a reason to add this?

Good catch! Reasoning was just a "I think this could be useful". But of course, we can read it later if we have a valid use case for it. Removed it for now.

marbetschar:
**Before**

Screenshot from 2020-10-31 12-31-54

**After**

Screenshot from 2020-10-31 12-48-15

marbetschar:
Changes are now saved to the backend when we edit a value in a PopoverEntry while not in full edit mode.

marbetschar:
@danrabbit regarding integration of Gtk.Revealer: We don't want to link its behaviour to internals of the PopoverEntry. In the Tasks app for example, we don't want the PopoverEntry to disappear when its value is deleted if we are in full edit mode.

As there is no way for the PopoverEntry to know the form is in full edit mode, I no longer think its a good idea to add it to the PopoverEntry directly as this may lead to issues or wrong assumptions down the road.

What do you think?

src/Widgets/DateTimePopover.vala
```diff
@@ -0,0 +1,116 @@
...
+
+        calendar.select_month (selected_datetime.get_month () - 1, selected_datetime.get_year ());
+        calendar.select_day (selected_datetime.get_day_of_month ());
+        timepicker.time = selected_datetime;
+
+        calendar.sensitive = timepicker.sensitive = true;
+    }
+
+    private void on_popover_closed () {
+        calendar.sensitive = timepicker.sensitive = false;
```
danrabbit, line 76: What's the reason for making these widgets insensitive while the popover is closed?

src/Widgets/EntryPopover.vala
```diff
@@ -0,0 +1,134 @@
...
+
+    construct {
+        events |= Gdk.EventMask.ENTER_NOTIFY_MASK
+            | Gdk.EventMask.LEAVE_NOTIFY_MASK;
+
+        popover = new Gtk.Popover (popover_button);
+
+        popover_button = new Gtk.MenuButton () {
+            always_show_image = true,
+            label = (placeholder != null && placeholder.length > 0 ? placeholder : _("Set Value")),
```
danrabbit, line 50: I can't imagine a case where we ever want a button to say "Set Value". This should probably be a construction property instead and save translators the work for a string that will never be shown

src/Widgets/EntryPopover.vala
```diff
@@ -0,0 +1,134 @@
...
+* General Public License for more details.
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+public abstract class Tasks.EntryPopover<T> : Gtk.EventBox {
+    public Gtk.Image image { get; set; }
```
danrabbit, line 21: Thinking about it more, this should probably be a construction property/argument and it would be more convenient for it to just be an icon name than a Gtk.Image

src/Widgets/EntryPopover.vala
```diff
@@ -0,0 +1,134 @@
...
+            var value_formatted = value_format (value);
+            if (value_formatted == null) {
+                popover_button.label = (placeholder != null && placeholder.length > 0 ? placeholder : _("Set Value"));
+
+                if (delete_button_revealer.reveal_child) {
+                    Timeout.add (150, () => {
+                        delete_button_revealer.reveal_child = false;
+                        return GLib.Source.REMOVE;
+                    });
+                }
```
danrabbit, line 111: Can you explain what this is for?

danrabbit (changes requested):

src/Widgets/EntryPopover.vala
marbetschar: Sounds good!

src/Widgets/EntryPopover.vala
marbetschar: wasn't sure about the drawbacks if we just pass an icon name - but seems like I was overthinking this :)

src/Widgets/EntryPopover.vala
marbetschar: The Timeout only serves a cosmetic purpose: If the user clicks on the delete button, the placeholder gets displayed immediately. Therefore it felt nicer to me to wait a bit till this change is visually done - and only after this, start the revealer's hiding transition.

src/Widgets/DateTimePopover.vala
marbetschar: The sensitive flag was used to determine if the value was changed by the user or programatically to avoid endless loops. Will check if this is still needed.

marbetschar: no longer needed. removed.

src/Widgets/EntryPopover.vala
marbetschar: gave it another try without the Timeout. Feels no longer weird, therefore removed the Timeout.

marbetschar:
@danrabbit thank you so much for your review, some very good points in here :+1:

src/Widgets/EntryPopover.vala
```diff
@@ -0,0 +1,141 @@
...
+*
+* You should have received a copy of the GNU General Public
+* License along with this program; if not, write to the
+* Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor,
+* Boston, MA 02110-1301 USA
+*/
+
+public abstract class Tasks.EntryPopover<T> : Gtk.EventBox {
+    public Gtk.Popover popover { get; private set; }
+    public string? icon { get; set; }
```
danrabbit, line 22: this should be `icon_name` so that it's not confused as a GLib.Icon

src/Widgets/EntryPopover.vala
```diff
@@ -0,0 +1,141 @@
...
+        notify["placeholder"].connect (() => {
+            if (value_format (value) == null) {
+                popover_button.label = placeholder;
+            }
+        });
+
+        notify["icon"].connect (() => {
+            ((Gtk.Image) popover_button.image).set_from_icon_name (icon, Gtk.IconSize.BUTTON);
+            popover_button.always_show_image = icon != null;
+        });
```
danrabbit, line 103: Is there a case where these properties will change?

src/Widgets/TaskRow.vala
```diff
@@ -91,23 +88,53 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
         summary_entry_context.add_class (Gtk.STYLE_CLASS_FLAT);
         summary_entry_context.add_provider (taskrow_provider, Gtk.STYLE_PROVIDER_PRIORITY_APPLICATION);
 
-        var due_image = new Gtk.Image.from_icon_name ("office-calendar-symbolic", Gtk.IconSize.BUTTON);
+        due_datetime_popover = new Tasks.DateTimePopover () {
+            icon = "office-calendar-symbolic",
+            placeholder = _("Set Due")
```
danrabbit, line 93: This should probably be internal to DateTimePopover right? Like there's not a example where some instance of DateTimePopover will set something different

danrabbit (changes requested):
Getting really close, just a couple more questions :)

danrabbit:
Woo okay, I think I can only find one more thing. There's an issue with saving the selected time. It seems like no matter what I do it doesn't save the time

marbetschar:
@danrabbit you need to press `Enter` after modifying the time to force the `time_changed` signal on the `Granite.Widgets.TimePicker` to be emitted. I guess this is an issue with the time picker?

If not, we can also check on popover hide if the `timepicker.time` has changed and emit the signal manually there.

danrabbit:
@marbetschar can you report that and assign me and I'll look at it tomorrow?

marbetschar:
out of curiosity: what's the reason of hard-coded placeholder and icon_name values in DateTimePicker constructor? Is the general idea to provide less configuration abilities in order to avoid future maintenance headaches?

Because in this case its kinda obvious that "Set Due" is probably not the placeholder we want in another context (which does not exist for now)...

marbetschar:
one last thing: Currently the calendar Widget does not highlight the selected day. Is this a styling thing on Odin or is something missing?

marbetschar:
@danrabbit I guess I fixed the time_changed issue in the DateTimePopover. Please Re-Review. From what I've tested the Granite.TimePicker works correctly, so I will close the issue over there.

danrabbit (approved):
Hey sorry for dropping the ball on this! Day selection highlight should be fixed now in Greenfield

> out of curiosity: what's the reason of hard-coded placeholder and icon_name values in DateTimePicker constructor? Is the general idea to provide less configuration abilities in order to avoid future maintenance headaches?

Yeah just separation of concerns. Keeping all of the construction of the DateTimePopover together instead of some in one class and some in another class

marbetschar:
@danrabbit all good - thanks for merging!

Is there a better way to ensure the `popover.show` & `popover.closed` signal handlers for the parent class `EntryPopover` are fired after the child-class `DateTimePopover`? As you've probably seen, right now I'm using `GLib.Idle.add` - although this works, it feels like a workaround.

PR #161, opened 2020-10-29
build: Remove the vapi directory

tintou:
The support for libecal-1.2 has been removed since a long time now


marbetschar (approved):
Everything still works as expected - therefore I guess this is fine ;)

PR #162, opened 2020-10-30
Code style and CSS fixes

danrabbit:
Widget construction should happen in the construct block. We don't use this style of constructing widgets during property declaration anywhere else. As you can see here, doing it that way lead to a lot of things being scoped to the whole class that are only accessed during construct.

We can also use property bindings to make our property setters simpler and can do a one line getter with `{ get; private set; }`

There were a number of extra properties here. `always_show_image` only needs to be set if the button also contains text. `use_popover` only applies if you're construction a menu button from a model.

You should use `-symbolic` icon names for flat styled icons so that they pick up the text color. This is especially important for dark mode.

Speaking of text color, there appears to be a bug in Greenfield that's making the error text color not work as expected, so I have to investigate that

### DIFF
--- a/data/EntryPopover.css
+++ b/data/EntryPopover.css
@@ -1,5 +1,5 @@
 /*
-* Copyright 2019 elementary, Inc. (https://elementary.io)
+* Copyright 2020 elementary, Inc. (https://elementary.io)
 *
 * This program is free software; you can redistribute it and/or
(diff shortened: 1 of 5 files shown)


marbetschar (approved):
Looks good, thank you very much for the feedback!

Issue #133, opened 2020-09-16
Missing icon sizes


cassidyjames:
We just have 64px at the moment.

- [ ] 16
- [ ] 24
- [ ] 32
- [ ] 48
- [x] 64
- [ ] 128

cc @elementary/ux :wink:

hezral:
i'd love to help but is the icon design final?

danrabbit:
@hezral Not particularly. I mostly did it as a placeholder. If you have a better design please feel free to propose it :)

hezral:
Thanks, that changes the task then. Anyway, i see @andirsun already made a PR for this so i'll back down for now :)

PR #166, opened 2020-11-25
add missing icon sizes

andirsun:
fix : #133


danrabbit (changes requested):
Hey thanks for submitting this branch! It looks like these scaled icons are pixel fitted as you can see in the image here:

Screenshot from 2020-12-08 13 08 57@2x

Strokes shouldn't fall on half pixels. Also keep in mind that the size of this icon is intended to match sizes of `text-x-preview` from elementary icons

PR #174, opened 2020-11-26
LocationPopover

### LINKED ISSUE #170: Support location based reminders
## Problem

In iOS we are able to get reminded of a task at a specific location. It would be great to at least display and allow editing this information in our Tasks app.

## Proposal

Unfortunately the way this information is stored and synced via CalDAV is Apple specific. So I suggest we start here and add support for further clients as needed.

## Prior Art

There has already been worked on a LocationPopover in PR #157. The DateTimePopover has already been merged into master. So we can create a new PR and glue those two things together.

In terms of the data structure, the way Apple stores this information is like this:

```
BEGIN:VTODO
CREATED:20201105T123437Z
DTSTAMP:20201105T123513Z
LAST-MODIFIED:20201105T123512Z
STATUS:NEEDS-ACTION
SUMMARY:Location Alert
UID:9710D97C-49D1-4395-90DB-14EE17E0B77A
BEGIN:VALARM
ACTION:DISPLAY
DESCRIPTION:Reminder
TRIGGER;VALUE=DATE-TIME:19760401T005545Z
UID:F68410D3-C36C-434B-A6E0-625A75DAB8FC
X-APPLE-PROXIMITY:ARRIVE
X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS=Via Monte Ceneri 1\\n6802 R
 ivera\\nSwitzerland;X-APPLE-RADIUS=100;X-APPLE-REFERENCEFRAME=1;X-TITLE=Vi
 a Monte Ceneri 1 6802 Rivera Switzerland:geo:46.141813\,8.917549
X-WR-ALARMUID:F68410D3-C36C-434B-A6E0-625A75DAB8FC
END:VALARM
END:VTODO
```

marbetschar:
This PR contains an initial draft for the `LocationPopover`. This supersedes #157 and fixes #170. The code also adds support for an iOS/Apple specific location based alarm. ~~This has not yet been extensively tested though. Also, I wonder how this location based alarms should be stored to support Android as well.~~

**UPDATE:** The Location based reminders have been tested on iOS and they work as expected (on arrival as well as on departure). Because I'm not aware of any standard way of doing this on Android, Android support is not part of this PR. Refer to issue #176 for any further discussion.

This is how the new popover looks like:

Screenshot from 2020-11-26 10-30-12


src/Util.vala
```diff
@@ -134,4 +134,200 @@ namespace Tasks.Util {
...
+            longitude = geo.get_lon ();
+            latitude = geo.get_lat ();
+        }
+
+        ICal.Property? apple_proximity_property = null;
+        ICal.Property? apple_location_property = null;
+
+        if (ecalcomponent.has_alarms ()) {
+            var all_alarms = ecalcomponent.get_all_alarms ();
+            foreach (var alarm in all_alarms) {
```
tintou, line 186:
```suggestion
            foreach (unowned var alarm in all_alarms) {
```

src/Util.vala
```diff
@@ -134,4 +134,200 @@ namespace Tasks.Util {
...
+
+        var geo_property_count = icalcomponent.count_properties (ICal.PropertyKind.GEO_PROPERTY);
+        for (int i = 0; i < geo_property_count; i++) {
+            var remove_prop = icalcomponent.get_first_property (ICal.PropertyKind.GEO_PROPERTY);
+            icalcomponent.remove_property (remove_prop);
+        }
+
+        if (ecalcomponent.has_alarms ()) {
+            var all_alarms = ecalcomponent.get_all_alarms ();
+            foreach (var alarm in all_alarms) {
```
tintou, line 280:
```suggestion
            foreach (unowned alarm in all_alarms) {
```

src/Location.vala
```diff
@@ -0,0 +1,33 @@
...
+    double latitude;
+    int accuracy;
+
+    Tasks.LocationProximity proximity;
+}
+
+public enum Tasks.LocationProximity {
+    ARRIVE,
+    DEPART
+}
```
tintou, line 33: You can even go the Vala way and define this:
```suggestion
public enum Tasks.LocationProximity {
    ARRIVE,
    DEPART;
    public static Tasks.LocationProximity from_string (string val) {
        if (val == "DEPART") {
            return Tasks.LocationProximity.DEPART;
        } else {
            return Tasks.LocationProximity.ARRIVE;
        }
    }
    public unowned string to_string () {
        switch (val) {
            case Tasks.LocationProximity.DEPART:
                return "DEPART";
            case Tasks.LocationProximity.ARRIVE:
            default:
                return "ARRIVE";
        }
    }
}
```
That way you can simply do afterward:
```vala
unowned string proximity_string = apple_proximity_property.get_value ().get_string ();
proximity = Tasks.LocationProximity.from_string (proximity_string);
```

src/Util.vala
```diff
@@ -134,4 +134,200 @@ namespace Tasks.Util {
...
+    }
+
+
+    //--- Location ---//
+
+
+    public Tasks.Location? get_ecalcomponent_location (ECal.Component ecalcomponent) {
+        var icalcomponent = ecalcomponent.get_icalcomponent ();
+
+        string? description = icalcomponent.get_location ();
```
tintou, line 168:
```suggestion
        unowned string? description = icalcomponent.get_location ();
```

src/Util.vala
```diff
@@ -134,4 +134,200 @@ namespace Tasks.Util {
...
+        }
+        return property;
+    }
+
+
+    //--- Location ---//
+
+
+    public Tasks.Location? get_ecalcomponent_location (ECal.Component ecalcomponent) {
+        var icalcomponent = ecalcomponent.get_icalcomponent ();
```
tintou, line 166:
```suggestion
        unowned ICal.Component? icalcomponent = ecalcomponent.get_icalcomponent ();
```

src/Util.vala
```diff
@@ -134,4 +134,200 @@ namespace Tasks.Util {
...
+                }
+            }
+
+            return location;
+        }
+        return null;
+    }
+
+    public void set_ecalcomponent_location (ECal.Component ecalcomponent, Tasks.Location? location) {
+        var icalcomponent = ecalcomponent.get_icalcomponent ();
```
tintou, line 269:
```suggestion
        unowned ICal.Component? icalcomponent = ecalcomponent.get_icalcomponent ();
```

marbetschar:
@tintout thanks for your review! I made the suggested changes, but now I get strange `Pango-WARNINGS` complainging about an invalid UTF-8 string whenever I enter a location name with a special character. Do you have any idea why?

For example, If I set the location to `Anthéor, France`, the values saved to the backend look like this:

```
BEGIN:VTODO
UID:0be972a2301dc88d717f4853c314aeaee5ea509c
DTSTAMP:20201207T200420Z
SUMMARY:France
CREATED:20201207T200526Z
LAST-MODIFIED:20201207T200607Z
LOCATION:Anthéor\, France
GEO:43.431563;6.886668
BEGIN:VALARM
X-EVOLUTION-ALARM-UID:95d12efc4b848ba72f08a171e04551d2888668ec
ACTION:DISPLAY
TRIGGER;RELATED=START:PT0S
X-APPLE-PROXIMITY:DEPART
X-APPLE-STRUCTURED-LOCATION;VALUE=URI;X-ADDRESS="Anthéor, France";X-APPLE-
 RADIUS=100;X-APPLE-REFERENCEFRAME=1;X-TITLE="Anthéor, France":geo:43.4315
 63\,6.886668
END:VALARM
END:VTODO
```

... this is the message I get (I added a debug statement in LocationPopover to make it more obvious)...

```
** (io.elementary.tasks:25881): DEBUG: 21:06:09.042: LocationPopover.vala:104: search_entry.text = '\xe0\x8a\xfe\xd4\u000bV'

(io.elementary.tasks:25881): Pango-WARNING **: 21:06:09.043: Invalid UTF-8 string passed to pango_layout_set_text()
```

And this is what is displayed on the screen:

Screenshot from 2020-12-07 21-06-12

marbetschar:
@tintou fixed it. But now I'm totally confused when its correct to use `var` and when its better to use `unowned` :laughing:

marbetschar:
@tintout I noticed the difference below between a `get_value ().get_string ()` and a `get_value_as_string ()` call. Does this need to be logged upstream as issue?

```
var apple_proximity_property_value = apple_proximity_property.get_value ().get_string ();
// -> apple_proximity_property_value = "                                                                                                                                      ";

var apple_proximity_property_value = apple_proximity_property.get_value_as_string ();
// -> apple_proximity_property_value = "DEPART";
```
Fixed in https://github.com/elementary/tasks/pull/174/commits/c0a11ac5fa70eb1ad036bdabc4ad0e3fe68c1b34

src/Widgets/LocationPopover.vala
```diff
@@ -0,0 +1,248 @@
...
+
+        search_entry = new Gtk.SearchEntry () {
+            placeholder_text = _("John Smith OR Example St."),
+            hexpand = true
+        };
+
+        var grid = new Gtk.Grid () {
+            margin = 6,
+            row_spacing = 6,
+            column_spacing = 6
```
danrabbit, line 69: As in Dialogs, margins in popovers should be 12px, and since there are no columns in this layout we don't need to set column spacing:
```suggestion
            margin = 12
            row_spacing = 12
```

src/Widgets/LocationPopover.vala
```diff
@@ -0,0 +1,248 @@
...
+        };
+
+        grid.attach (map_frame, 0, 0);
+        grid.attach (search_entry, 0, 1);
+        grid.attach (location_mode, 0, 2);
```
danrabbit, line 74: I wonder if the search entry should be up top since information flows from the search to the map

src/Widgets/LocationPopover.vala
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    construct {
+        map_embed = new GtkChamplain.Embed () {
+            height_request = 120,
+            width_request = 220
+        };
+
+        point = new Marker () {
+            draggable = true
```
danrabbit, line 42: It doesn't seem like the location information updates when I drag the marker, so maybe it shouldn't be draggable for now?

danrabbit (changes requested):
Right on, this is pretty slick. Couple of inline comments.

When I set a location reminder in iOS and then open that in Tasks, the map doesn't zoom to the location I've supplied

It would probably be helpful to show "Arriving: 123 Foo St" or "Departing: 456 Bar Ave" depending on the selection

marbetschar:
@danrabbit good points! After implementing your feedback, the `LocationPopover` now looks like this:

Screenshot from 2020-12-09 19-11-21

marbetschar:
@danrabbit good catch regarding the map issue!

Apple uses two parameters to store the location as text: `X-ADDRESS` and `X-TITLE`. Where `X-TITLE` contains the location which should be displayed to the user - which is not necessarily the postal address. It depends on how you stored this piece of information in your address book. On the other hand, the `X-ADDRESS` parameter seems to always contain the postal address.

As we can't really make use of `X-TITLE` for now, I changed the code to solely respect `X-ADDRESS`. Also I've figured out some ways to make the coordinate parsing and set parameter code blocks simpler.

src/Widgets/LocationPopover.vala
```diff
@@ -0,0 +1,245 @@
...
+    private GtkChamplain.Embed map_embed;
+    private Gtk.SearchEntry search_entry;
+    private GLib.Cancellable search_cancellable;
+    private Granite.Widgets.ModeButton location_mode;
+    private Marker point;
+
+
+    public LocationPopover () {
+        Object (
+            icon_name: "location-inactive-symbolic",
```
danrabbit, line 30: For this we probably want `mark-location-symbolic` since the location-active/inactive icons are meant to indicate the active status of location services and not that this object is a location
```suggestion
            icon_name: "mark-location-symbolic",
```

danrabbit (commented):
Small inline suggestion

danrabbit (commented):
I've got 403 forbidden when trying to add any location data, is it expected that this is read only?

I also still don't have location data showing for my task which has a title and geo but no address

But I can confirm that locations that have a title now show the title in the task list and maps show the correct location and the address is shown in the search entry

I wonder if we should make this popover a little larger since the search entry doesn't seem to accomodate the width of a whole street address with postal codes and everything

marbetschar:
> I've got 403 forbidden when trying to add any location data, is it expected that this is read only?

No, this is not expected. Can you please elaborate if/when you exactly get a 403 along with any additional information available? (E.g. which backend you are using (I mainly test against Nextcloud))

> I also still don't have location data showing for my task which has a title and geo but no address

Can you please re-test? The previously modified tasks may contain corrupted geo information.This is now fixed: The `X-APPLE-STRUCTURED-LOCATION` property should contain a url formatted value like `geo:47.363875\,8.556419`. I wrongly assumed I'd be able to read/write this value using `ICal.Value.Geo` - but according to the [CalDAV specs](https://icalendar.org/iCalendar-RFC-5545/3-8-1-6-geographic-position.html), the Geo value is expected to be semicolon-separated. The solution implemented now does manually parse and update the property value.

If this still does not work: Can you provide the *.ics file containing the tasks?

> I wonder if we should make this popover a little larger since the search entry doesn't seem to accomodate the width of a whole street address with postal codes and everything

Made the LocationPopover a tad larger.

danrabbit:
@marbetschar I'm using the old iCloud backend. If I modify the location in any way and close the popover I get:
`** (io.elementary.tasks:39405): WARNING **: 14:30:29.891: TaskModel.vala:495: Cannot modify calendar object: Failed to put data: HTTP error code 403 (Forbidden)`

For what it's worth, I get this in Evolution too. I filed an issue about making sure we throw these errors as a dialog, but if there's a way to know that we're read only it would be nice to just make this insensitive ahead of time

It looks like the issue with the location title no address thing is fixed. Nice job!

danrabbit (approved):
This works for me for reading tasks and afaict, works no worse than Evolution for writing. So I'm happy to merge if you're happy and we can always iterate if necessary

marbetschar:
I created a dedicated issue for this iCloud issue, so we cap keep track of it there.

PR #177, opened 2020-12-07
Make description textview useable

### LINKED ISSUE #173: Make description textview "usable"
## What Happened

Although the `description_textview` in `Tasks.TaskRow` works, it is not very convenient to use. Especially if you edit a larger text.

Things that don't work atm:

- Place the cursor at an arbitrary position
- Select a range of characters

## Expected Behavior

Nothing special, the textview should simply behave like we are used to from other applications and therefore allow for easy editing.

marbetschar:
This is an attempt to fix #173. It seems to work a tiny bit better now. But for whatever reason every second click the `description_textview` looses its focus. So in order to set the cursor somewhere else we need to click twice - which is still very annoying.


alainm23:
@marbetschar Try set `can_focus = false;` to `Tasks.TaskRow`.

marbetschar:
thanks for the hint @alainm23 - it worked!

danrabbit (approved):
Seems to work as expected :+1:

Issue #141, opened 2020-09-30
Fill in appdata


ryonakano:
It has no summary or description atm:

https://github.com/elementary/tasks/blob/70fb4c8aeeee74e942927f2fba56ef8428134af0/data/tasks.appdata.xml.in#L8-L11

We need to fill in these texts before production release

andirsun:
How can provide me the text to help with PR?

PR #181, opened 2020-12-09
Fill in appdata

trst:
Addresses #141 

Took a quick crack at the copy. Super open to feedback!

Summary for tasks was taken from the elementary/tasks repository tagline/description/note (or whatever its called). I looked to the Gnome Tasks for a starting point on the Description, though I ended up going a different direction.


danrabbit (changes requested):
Hey thanks for working on this!

This seems like a pretty good starting point. The only thing is we try to steer away from mentioning elementary OS in appdata since every app in AppCenter is made for elementary OS, so this isn't really unique and doesn't describe a differentiating feature of the app

trst:
Fair point. Is it descriptive enough with that part of the string removed? Or should I give it a rethink?

danrabbit:
Yeah I think that's a good starting point and we can always iterate and add more later :)

Only thing is to remove the elementary OS mention from the summary label and we're good to go!

trst:
Doh! Should be sorted, or I’m going to eat my hat 😂

danrabbit (approved):
Looks good! Thanks for working on this :)

PR #182, opened 2020-12-10
TaskRow: Prevent the use of NULL ICal.Time object

tintou:
The ICal.Component might exist but not the created time.


marbetschar (approved):
Good catch! Works as expected - thank you very much!

PR #187, opened 2021-02-04
Change list color in EDS and WebDAV backend

### LINKED ISSUE #167: WebDAV: Save task list color to backend (e.g. Nextcloud)
## What Happened

If we change the task list color, this change should also be reflected in the corresponding backend. Unfortunately [libevolution does not provide any high level API](https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/261#note_944689) to do so, so we need to use specific, lower level API's for each backend.

## Expected Behavior

The task list color is saved locally as well as on the server.

## Existing work

For WebDAV this has already been implemented for the task list's `display_name`. We basically do the same steps for the color, but need to use ["calendar-color"](https://gitlab.gnome.org/GNOME/evolution/-/blob/master/src/e-util/e-webdav-browser.c#L1706) instead of ["displayname"](https://gitlab.gnome.org/GNOME/evolution/-/blob/master/src/e-util/e-webdav-browser.c#L1691) for the `E.WebDAVPropertyChange`. IMHO the way to go is to create a new function `update_task_list_color` and implement this behaviour there:

https://github.com/elementary/tasks/blob/0a11aefa6d497f02555d674463bd90bbd1587cdd/src/TaskModel.vala#L151-L191

marbetschar:
This PR fixes #167 by updating the list color property in the WebDAV backend as well as in the local EDS.

Changing the color property only in the local EDS turned out to not be reflected in the WebDAV backend. Technically it would be enough, to just update the color in the WebDAV backend. However, changing it only there turned out to take quite a while until the change is visible in the UI. Hence I'm updating it locally first to immediately reflect the change in the UI - and then send an update request to the WebDAV server.

This way, the UI feels more responsive.


danrabbit:
Sorry for taking so long to review this! Can you resolve conflicts here and then ping me for review :)

danrabbit (commented):
I can't confirm that this works. The color is changing locally as expected, but the changes aren't showing when I sync tasks to my other computer

marbetschar:
Added the following changes:

- Throw more specific errors in TaskModel's update_task_list_color where needed:
    - `Tasks.TaskModelError.BACKEND_ERROR` in case the backend is not supported (eg. gtasks)
    - `Tasks.TaskModelError.INVALID_ARGUMENT` in case we don't have enough information to actually do the color change
- Reset the local task list color in case of any backend error
    - this made it necessary to reset the selected color button, thus introducing the `update_selected_task_list_color` method
- Made sure we only send a request for color change if there was a modification in the UI:
    - color_button needs to be active (the toggle signal is emitted for deactivation too)
    - The color itself must be different than the one set currently (this avoids sending a color update request to the backend on initial load)

danrabbit (changes requested):
I can confirm that I'm getting errors now which is good! But I'm getting a `null` backend message for iCloud/Caldav and for local:

```
** (io.elementary.tasks:10753): CRITICAL **: 12:02:34.555: tasks_task_model_get_collection_backend_name: assertion 'source != NULL' failed

(io.elementary.tasks:10753): GLib-CRITICAL **: 12:02:34.557: g_utf8_strdown: assertion 'str != NULL' failed
```

And it looks like the actual color is reset, but the selection in the settings menu is not:

Screenshot from 2021-03-12 12 04 24@2x

src/Widgets/ListSettingsPopover.vala
```diff
@@ -234,4 +219,83 @@ public class Tasks.ListSettingsPopover : Gtk.Popover {
...
+                select_task_list_color (source);
+                dialog_update_task_list_color_error (e);
+            }
+        });
+    }
+
+    private void dialog_update_task_list_color_error (Error e) {
+        string error_message = e.message;
+
+        GLib.Idle.add (() => {
```
danrabbit, line 287: I don't see that we're using threads here, so it doesn't seem like an idle is necessary

marbetschar: Interesting: It works without `GLib.Idle`...

Does this mean `GLib.Idle` is not needed for async operations? Asking because coming from iOS, there we also need to dispatch UI work explicitly to the main thread in case work was done asynchronously.

marbetschar:
> I can confirm that I'm getting errors now which is good! But I'm getting a `null` backend message for iCloud/Caldav and for local:
> 
> ```
> ** (io.elementary.tasks:10753): CRITICAL **: 12:02:34.555: tasks_task_model_get_collection_backend_name: assertion 'source != NULL' failed
> 
> (io.elementary.tasks:10753): GLib-CRITICAL **: 12:02:34.557: g_utf8_strdown: assertion 'str != NULL' failed
> ```
> 
> And it looks like the actual color is reset, but the selection in the settings menu is not:

Both fixed in latest commit.

src/Widgets/ListSettingsPopover.vala
danrabbit: @tintou would be able to explain this better, but I think what you're talking about is wrapping something in a [GLib.MainLoop](https://valadoc.org/glib-2.0/GLib.MainLoop.html) which afaik you only need to do if you're using [GLib.Thread](https://valadoc.org/glib-2.0/GLib.Thread.html)

marbetschar:
@danrabbit anything missing to get this merged?

danrabbit (approved):

PR #188, opened 2021-02-04
WebDAV: Delete Task List

marbetschar:
This PR adds the functionality to delete Tasks Lists if they are either stored locally or in a WebDAV backend.


src/TaskModel.vala
```diff
@@ -201,6 +201,42 @@ public class Tasks.TaskModel : Object {
         }
     }
 
+    public async void remove_task_list (E.Source task_list) throws Error {
+        var registry = get_registry_sync ();
+        var task_list_extension = (E.SourceTaskList) task_list.get_extension (E.SOURCE_EXTENSION_TASK_LIST);
```
danrabbit, line 206: Getting a compiler warning about this:
```
../src/TaskModel.vala:206.13-206.31: warning: local variable `task_list_extension' declared but never used
```

danrabbit (changes requested):
I can confirm this works for local backend, just one comment

src/TaskModel.vala
marbetschar: Good catch! Removed the var.

danrabbit (approved):

PR #189, opened 2021-02-12
Show dialog if we fail to carry out the user's intent

### LINKED ISSUE #180: Don't fail silently when unable to write
https://github.com/elementary/tasks/blob/4a66b7e3ddf6ec8d52c8196eac658c80756bb122/src/TaskModel.vala#L495

It looks like currently we're failing silently when we're unable to write to the calendar object. This should be an error dialog since it's a failure to carry out the user's intent

marbetschar:
Fixes #180. If we agree on this concept, we can expand it to task lists - which would save us some lines of code.


danrabbit (changes requested):
I don't think we should rely on the library error output here to provide feedback for users. The problem with abstracting error dialogs like this is that they're so generic that the feedback isn't useful for resolving the problem. 

For example in `add_task_async ()` the user is getting the exact same error for when we're unable to get the client as when we're unable to create an object. We could be providing much more accurate feedback than just that we failed to add a new task

marbetschar:
Valid point. What about redefining `error_received` to something like this (not sure whether if its good design to have translations within TaskModel or not):

```
public signal void error_received (string title, string detail, Error? error = null);
```

Another way would be returning more detailed errors extending the already existing `errordomain Tasks.TaskModelError` - but that might be a bit overkill?

danrabbit:
I'm not sure that this needs to be further abstracted at all. Granite.MessageDialog is already an abstraction

marbetschar:
@danrabbit how do you suggest to transport errors and their context from TaskModel to the UI? Should TaskModel itself be responsible of showing the MessageDialog?

Till now I thought of TaskModel as being part of the Model Layer and therefore UI agnostic … but maybe that's overthinking stuff…?

danrabbit:
@marbetschar public methods in the taskmodel should probably throw an error and then we can handle the error from wherever the method is called

marbetschar:
@danrabbit thanks for your feedback! The involved part of the code seems much cleaner now: All task CRUD methods are `async` / non-blocking and potentially errors thrown are returned to the UI.

Although there might be better detail descriptions, I'm quite happy with the current state of work.

danrabbit (approved):

PR #192, opened 2021-02-16
Sort Task Lists using calendar-order

### LINKED ISSUE #169: Support user defined sorting of Task Lists
I'm not sure about GTasks, but at least for WebDAV there is a way to rearrange task lists in the backend. Instead of just sorting them alphabetically in our Tasks app, we should reflect the order created by the user.

For WebDAV there has to be a property we can use for sorting the task lists. If this is not present, we can still fall back to alphabetical sorting.

marbetschar:
This PR will eventually fix #169 by adding support for the `calendar-order` property. Because this property was added in a newer Evolution Data Server version than shipped with the Ubuntu 20.04 LTS base, using Flatpak is a must.

This way, we can package a newer version of the needed `evolution-data-server` and ship it within the Tasks app.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 *~
 build/*
+.flatpak-builder/
--- a/README.md
+++ b/README.md
@@ -3,6 +3,33 @@
 
 ## Building and Installation
 
+### Build with Flatpak
+
+_Starting with elementary 6 Odin, Flatpak is the preferred build method._
+
+You'll need to install the following dependencies:
+
+```bash
+flatpak --user install flathub \
+  org.gnome.Sdk//3.38 \
+  io.elementary.BaseApp//juno-20.08
+```
+
+Run `flatpak-builder` to build:
+
+```bash
+flatpak-builder --force-clean build io.elementary.tasks.yml
+```
+
+To install, use `flatpak-builder --install`, then execute with `io.elementary.tasks`:
+
+```bash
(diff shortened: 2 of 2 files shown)


marbetschar:
The necessary code is in place, but right now throws `error: `E.util_source_compare_for_sort' is not available in libedataserver-1.2 3.39.2. Use libedataserver-1.2 >= 3.40`.

Therefore we need to wait for the EDS 3.40 release and adjust the `io.elementary.tasks.yml` accordingly.

Marukesu:
> Therefore we need to wait for the EDS 3.40 release and adjust the `io.elementary.tasks.yml` accordingly.

A problem can be the DBus interfaces, we need them in the same version of the host, since we use the online accounts plug to login.

marbetschar:
@Marukesu I merged EDS v3.40 and `E.util_source_compare_for_sort` gets executed in `MainWindow.vala:323`. But it does not seem to have any effect.

How can I verify if this is related to the locally installed EDS version (not the one in the Flatpak) not populating the `calendar-order` field?

marbetschar:
Alright; the [`E.SourceSelectable.order`](https://valadoc.org/libedataserver-1.2/E.SourceSelectable.html) field is always `0` - so the new EDS function `E.util_source_compare_for_sort` uses the `display_name` for sorting.

Therefore this PR is ready to be merged. However, if we really want to make use of the new `calendar-order` property, the non-Flatpak version of EDS needs to be updated.

PR #193, opened 2021-02-16
Fix compile errors in EDS v3.39.2+

marbetschar:
And provide backwards compatibility to the shipped eds version in the Ubuntu 20.04 LTS as well.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -34,8 +34,9 @@ tasks_deps = [
     dependency('libgdata')
 ]
 
-if libecal_dep.version().version_compare('<=3.39')
-    add_project_arguments('--define=EDS_3_39', language: 'vala')
+if libecal_dep.version().version_compare('>=3.39.2')
+    add_project_arguments('--define=HAS_EDS_3_40', language: 'vala')
+else
     tasks_deps += meson.get_compiler('vala').find_library('libecal-2.0-fixes', dirs: meson.current_source_dir() / 'vapi')
 endif
 
--- a/meson.build
+++ b/meson.build
@@ -35,6 +35,7 @@ tasks_deps = [
 ]
 
 if libecal_dep.version().version_compare('<=3.39')
+    add_project_arguments('--define=EDS_3_39', language: 'vala')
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -35,6 +35,7 @@ tasks_deps = [
 ]
 
 if libecal_dep.version().version_compare('<=3.39')
+    add_project_arguments('--define=EDS_3_39', language: 'vala')
     tasks_deps += meson.get_compiler('vala').find_library('libecal-2.0-fixes', dirs: meson.current_source_dir() / 'vapi')
 endif
```
tintou, line 40: We usually do the reverse for readability
```suggestion
if libecal_dep.version().version_compare('>=3.39.2')
    add_project_arguments('--define=HAS_EDS_3_40', language: 'vala')
else
    tasks_deps += meson.get_compiler('vala').find_library('libecal-2.0-fixes', dirs: meson.current_source_dir() / 'vapi')
endif
```

marbetschar: wow, that was quick :)

makes sense, will implement as advised!

src/TaskModel.vala
```diff
@@ -272,12 +276,11 @@ public class Tasks.TaskModel : Object {
...
                         webdav_server_uri = new Soup.URI (webdav_discovered_source.href.dup ());
                     }
-                    /**
-                     * TEMPORARY WORKAROUND: `E.webdav_discover_do_free_discovered_sources`
-                     * Remove this line, once the following commit of libedataserver is released:
-                     * https://gitlab.gnome.org/GNOME/evolution-data-server/-/commit/9d1505cd3518ff32bd03050fd898abf89d31d389
-                     */
+
+#if HAS_EDS_3_40
+#else
```
tintou, line 281:
```suggestion
#if !HAS_EDS_3_40
```

marbetschar: wasn't aware this is possible - thanks for the hint!

tintou (approved):

PR #197, opened 2021-02-23
Keep state across lists

### LINKED ISSUE #151: Create multiple instances for each Task List.
## Prerequisites

## Feature
What if I want to copy information between tasks, this is not possible. Currently every time the user switches between task lists the current task view is eliminated and replaced by the new tasks, it is a quick solution but not very UX friendly. 

- Create multiple instances for each task list.
- I [fixed](https://github.com/alainm23/planner/blob/master/src/MainWindow.vala#L539) this problem in Planner.

**Additional context**
task

marbetschar:
This PR fixes #151 by adding a `Gtk.Stack` to switch between the different task list views. This allows for keeping the current state of a list view and its tasks.


src/MainWindow.vala
```diff
@@ -219,8 +219,32 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+
+                task_lists.foreach ((source) => {
+                    E.SourceTaskList list = (E.SourceTaskList)source.get_extension (E.SOURCE_EXTENSION_TASK_LIST);
+
+                    if (list.selected == true && source.enabled == true) {
+                        add_source (source);
+
+                        if (last_selected_list == "" && default_task_list == source) {
+                            listbox.select_row (source_rows[source]);
+                            listbox.row_selected (source_rows[source]);
```
danrabbit, line 240: `row_selected` should be emitted automatically when `select_row` is called. I don't think you need to forcefully emit it like this

src/MainWindow.vala
```diff
@@ -155,18 +155,33 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
-                        Tasks.Application.settings.set_string ("selected-list", "scheduled");
+                        var scheduled_uid = "scheduled";
+
+                        listview = (Tasks.ListView) listview_stack.get_child_by_name (scheduled_uid);
+                        if (listview == null) {
+                            listview = new Tasks.ListView (null);
+                            listview_stack.add_named (listview, scheduled_uid);
+                        }
+
+                        listview.remove_views ();
```
danrabbit, line 184: What's this about? If the only time we're doing this now is when we add a scheduled list, shouldn't that just be a part of the way the scheduled list is constructed?

src/MainWindow.vala
```diff
@@ -155,18 +155,33 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+                        }
 
+                        listview_stack.set_visible_child_name (source_uid);
+                        Tasks.Application.settings.set_string ("selected-list", source_uid);
                         ((SimpleAction) lookup_action (ACTION_DELETE_SELECTED_LIST)).set_enabled (source.removable);
 
                     } else if (row is Tasks.ScheduledRow) {
-                        listview.source = null;
-                        Tasks.Application.settings.set_string ("selected-list", "scheduled");
+                        var scheduled_uid = "scheduled";
```
danrabbit, line 176: perhaps this should be a constant since it is referenced in a few other places like saving and restoring the selection

danrabbit (changes requested):
I suppose this will increase our memory footprint, but really the best way to solve the issue. Left a few questions and comments.

I suppose that since `ListView` is now not just a view that is updated but a widget which has many instances it should be moved to `widgets` :)

src/MainWindow.vala
marbetschar: Good point! Added the `SCHEDULED_LIST_UID` constant.

src/MainWindow.vala
marbetschar: Removed the three calls - but now the Scheduled list is broken. To reproduce:

1. Compile from this branch and start Tasks
2. Select the Scheduled list
3. Exit the Tasks app
4. Start the Tasks app, now it looks like this:

Screenshot from 2021-02-24 09-55-03

... is this, because the "Scheduled" list is already selected on startup and therefore the `listbox.row_selected.connect` is never called? If so, how do you suggest to fix this?

src/MainWindow.vala
marbetschar: Yeah, that's a bit ugly. The reasoning for this is:

We need to create a view (read "query") for each source, so we can display all scheduled tasks across all sources on the scheduled list. But since sources themselves can be created, deleted or deactivated and because I also added the scheduled list to the `Gtk.Stack` it is no more enough doing this just once on constructing the scheduled list.

Therefore I'm removing any potential existing views (queries), when the user sets the visible child of the stack to the scheduled list, then add all views for currently existing and active sources, then displaying the scheduled list.

Other potential solutions for this problem:
- Remove scheduled list from `Gtk.Stack` and fully re-construct it whenever the user switches to the scheduled list
- Keep scheduled list in `Gtk.Stack`, but listen to source crud events (this seemed to cause significant refactoring)

src/MainWindow.vala
marbetschar: Fixed this by removing all `row_selected` apart from the one for the Scheduled list - don't know why its needed there, but now it works as intended.

marbetschar:
> I suppose this will increase our memory footprint, but really the best way to solve the issue. Left a few questions and comments.
> 
> I suppose that since `ListView` is now not just a view that is updated but a widget which has many instances it should be moved to `widgets` :)
> 
> It might be too much for this branch (or maybe it actually makes things easier), but I imagine the scheduled view could be its own widget and that would take out a lot of conditional statements

Created https://github.com/elementary/tasks/issues/201 - will take care of refactoring in a next PR.

danrabbit (approved):
Works as expected and further refactoring is planned for further branches so it's a :+1: from me :) Nice work

Just needs conflicts resolved

PR #198, opened 2021-02-23
Added support for X-APPLE-SORT-ORDER

marbetschar:
This PR lays the foundation to enable us to manually sort tasks within a list.

Unfortunately the VTODO RFC does not support a "standard" way of sorting. However, this has been solved on Apple devices by introducing the custom field `X-APPLE-SORT-ORDER`.

If the custom field is not available, we calculate the sort value the same way Apple does: The number of seconds between the creation of the task and 20010101T000000Z (Cocoa/Webkit epoch).

See the Nextcloud Tasks app implementation of `X-APPLE-SORT-ORDER` for further details and a discussion about pros/cons:

- https://github.com/owncloud/tasks/pull/358
- https://github.com/owncloud/tasks/issues/86


src/Util.vala
```diff
@@ -166,23 +166,60 @@ namespace Tasks.Util {
...
+                return x_property;
+            }
+        }
+        return null;
+    }
+
+
+    //--- Sort ---//
+
+
```
danrabbit, line 197: Seems like a little bit of extra whitespace and a quite obvious comment :)

src/Util.vala
```diff
@@ -166,23 +166,60 @@ namespace Tasks.Util {
...
+    public string get_ecalcomponent_sort (ECal.Component ecalcomponent) {
```
danrabbit, line 198: Maybe this should more explicitly be something like `get_apple_sortorder` to explain why it's a string

src/Util.vala
```diff
@@ -166,23 +166,60 @@ namespace Tasks.Util {
...
+            * if the property is not set, Apple uses the
+            * number of seconds between the creation of
+            * the task and 20010101T000000Z (Cocoa/Webkit epoch)
+            * as value for sorting.
+            */
+            sort = ecalcomponent.get_created ()
+            .subtract (new ICal.Time.from_string ("20010101T000000Z"))
+            .as_int ()
+            .to_string ();
+        }
```
danrabbit, line 221: Hrm, it seems weird to convert to string here and it seems the only reason that's being done is because the value for `X-APPLE-SORT-ORDER` is a string. It might be better to have this function only do one thing (get the apple sort order) and handle the null case in the listbox sort function still

src/Util.vala
marbetschar: True :laughing: - removed.

src/Util.vala
marbetschar: Changed to `public string? get_apple_sortorder_property_value`

src/Util.vala
marbetschar: Created a new `public ICal.Duration get_apple_sortorder_default_value` in `Tasks.Util`, which returns the default value. Converting to the desired format is then done in the listbox sort function.

src/Util.vala
```diff
@@ -166,23 +166,55 @@ namespace Tasks.Util {
...
+
+    /**
+    * if the X-APPLE-SORT-ORDER property is not set,
+    * Apple uses the number of seconds between the creation of
+    * the task and 20010101T000000Z (Cocoa/Webkit epoch)
+    * as value for sorting.
+    */
+    public ICal.Duration get_apple_sortorder_default_value (ECal.Component ecalcomponent) {
+        return ecalcomponent.get_created ().subtract (new ICal.Time.from_string ("20010101T000000Z"));
     }
```
danrabbit, line 218: What's the benefit of this over using the old comparison method? I'm not a huge fan of magic numbers

marbetschar: I'm not a huge fan of this either - but in `ListView.sort_function`, we need something to compare against, in case `sortorder_a` is set but `sortorder_b` isn't: If we use the values as before, all tasks with `dtstart` or `dtstamp` set (but no `x-apple-sort-order`) will be shown before the manually sorted ones (the ones with `x-apple-sort-order` set) - therefore mainly defeating the purpose of manual sorting.

So either I'm missing something here, or your question boils down to whether we should implement custom sorting or not. And to be more specific here: The Apple way of custom sorting.

Because there is no official way for custom sorting in the CalDAV standard, we need to come up with our own solution - if we want to provide this feature. The Apple method for custom sorting has the added benefit by already being implemented in a bunch of software such as Nextcloud, iOS (and potentially others).

marbetschar: FWIW: I tested with using `row_a.task.get_created ().as_timet ().to_string ()` instead of the Apple magic value in case the property value is not available. While technically it works, it leads to different results than on the iOS device. However, we can go this route and re-implement the Apple default behavior again later on if there is an actual need for it?

danrabbit: Ah okay, I think I just wasn't understanding how all these pieces fit together. Maybe can you amend the comment with a brief explanation that we're doing this so that we can match the sort order on iOS devices?

danrabbit (approved):

Issue #171, opened 2020-11-25
Trigger data refresh on app start


marbetschar:
## What Happened

It seems like Tasks needs quite a while to pick up changes done in the backend. Therefore I suggest we trigger a data refresh on each app start Also it would be great to indicate this activity to the user by showing a spinner.

## Expected Behavior

Whenever the app is started, it should provide the user the latest information.

## Steps to Reproduce

1. Open the Tasks app and wait a bit
2. Close the Tasks app
3. Change a task in the backend
4. Re-Open the Tasks app

The task in question still shows the old values.

Issue #74, opened 2019-12-20
Sidebar icon doesn't update


danrabbit:
The sidebar icon doesn't seem to update once a connection has been established or while doing work

PR #200, opened 2021-03-02
Fix #172: Allow manual refresh with Ctrl+r

### LINKED ISSUE #172: Trigger manual data refresh with Ctrl+r
## Problem

It would be great to allow the user to manually refresh the data while the app is running.

## Proposal

This can be done with an accelerator action for `Ctrl+r`. We might also want to indicate refresh activity by showing a spinner animation (can be the same as in #171).

## Prior Art

Backend refresh triggers are already used after renaming a Task List. Maybe we can just use the same technique:

https://github.com/elementary/tasks/blob/0a11aefa6d497f02555d674463bd90bbd1587cdd/src/TaskModel.vala#L180

Also some accelerators (keyboard shortcuts) have already been implemented in `Tasks.MainWindow`, so we can use those to figure out how to create another one:

https://github.com/elementary/tasks/blob/0a11aefa6d497f02555d674463bd90bbd1587cdd/src/MainWindow.vala#L23

marbetschar:
This PR introduces a new `Ctrl + r` keyboard binding which allows to trigger a refresh of the task lists manually. This is useful whenever there were changes in the backend which are not reflected in the UI yet.

Also we may want to build upon this and trigger the refresh on app start if there is internet connection available (see #171)


src/MainWindow.vala
```diff
@@ -271,6 +274,46 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+        });
+    }
+
+    private void dialog_refresh_task_list_error (Error e) {
+        string error_message = e.message;
+
+        GLib.Idle.add (() => {
+            var error_dialog = new Granite.MessageDialog (
+                _("Refresh task list failed"),
+                _("The backend may be unavailable or unable to read from."),
```
danrabbit, line 303: We should probably try to be as accurate and helpful here as we have the information to be. It would be nice to use more natural sounding language as well. Something like:
```
Unable to get updates for "Groceries"
The tasks at "me@example.com" may be unavailable or unable to be read."
```

I wonder about that second line as well. If we didn't fail to refresh the collection, then is it really a possibility that the _backend_ is unavailable? Should this instead say that this task list may be unavailable?

src/MainWindow.vala
```diff
@@ -271,6 +274,46 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
     }
 
+    private void dialog_refresh_collection_error (Error e) {
+        string error_message = e.message;
+
+        GLib.Idle.add (() => {
+            var error_dialog = new Granite.MessageDialog (
+                _("Refresh collection failed"),
+                _("The backend may be unavailable or unable to read from."),
+                new ThemedIcon ("dialog-error"),
```
danrabbit, line 284: Maybe this should `preferences-desktop-online-accounts` with a badge of `dialog-error` to make it clearer this is an issue with getting data from somewhere else?

src/MainWindow.vala
```diff
@@ -271,6 +274,46 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
         });
     }
 
+    private void dialog_refresh_collection_error (Error e) {
+        string error_message = e.message;
+
+        GLib.Idle.add (() => {
+            var error_dialog = new Granite.MessageDialog (
+                _("Refresh collection failed"),
+                _("The backend may be unavailable or unable to read from."),
```
danrabbit, line 283: Is there a more user-facing way to phrase this? Maybe something like:
```
Unable to get updates from "me@example.com"
The tasks stored there may be unavailable or unable to be read from.
```

danrabbit (changes requested):

src/MainWindow.vala
```diff
@@ -271,6 +274,46 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
         });
     }
 
+    private void dialog_refresh_collection_error (Error e) {
+        string error_message = e.message;
+
+        GLib.Idle.add (() => {
```
danrabbit, line 280: What's the reason for wrapping this in an idle?

src/MainWindow.vala
```diff
@@ -271,6 +274,46 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
         });
     }
 
+    private void dialog_refresh_collection_error (Error e) {
+        string error_message = e.message;
```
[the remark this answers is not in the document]
marbetschar: I ran into problems where `e` was not available anymore when using `GLib.Ide`. But as it turned out `GLib.Idle` is not needed - therefore the copy here is not needed too.

src/MainWindow.vala
marbetschar: Changed the error message accordingly. Also, you're right - the backend is indeed available in this case so I changed the message to reflect this.

src/MainWindow.vala
marbetschar: Done.

src/MainWindow.vala
marbetschar: I (wrongly) assumed `GLib.Idle` is needed when we want to display something in the UI and this UI work is called from an async operation.

marbetschar:
@tintou is there a way to let a source update its connection status automatically and get notified about the change? Doing this manually based on user interaction seems wrong - and https://github.com/elementary/tasks/issues/74 suggests there should be a better way than this:

https://github.com/elementary/tasks/blob/6dbdb28dc01cced1c15f3ae126708178e7ac106e/src/MainWindow.vala#L334-L346

Marukesu:
Testing it with GTasks, sometimes the refesh seems to froze and sometimes it re-create all the GTaksks sources even the disabled ones.

marbetschar:
@Marukesu I don't seem to run into these problems. Can you  please describe how to reproduce?

Marukesu:
It happens randomly, sometimes this warning is shown on terminal:
```
(io.elementary.tasks:3240): libedbus-private-WARNING **: 16:07:09.721: Error setting property 'ConnectionStatus' on interface org.gnome.evolution.dataserver.Source: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.DBus.Properties” on object at path /org/gnome/evolution/dataserver/SourceManager/Source_24 (g-dbus-error-quark, 19)
```

Marukesu:
So, i got the bug reproducible by refreshing, changing the selected tasklist and refreshing again.

Also, it's looks like it's refreshing all tasklist, isn't better to only refresh the selected tasklist and let the eds auto-refresh update the rest?

marbetschar:
Since we have auto-refresh now upon task list selection, we no longer need this.

PR #203, opened 2021-03-16
GTasks: Support Add Task List

### LINKED ISSUE #164: Google Tasks: Add a Task List
## Problem

It should be possible to add a task list for a Google backend.

## Proposal

Since there is no high level API in EDS for renaming tasks lists in the backend, we need to find a lower level API way to do this. See how its done for a regular WebDAV backend in PR #150

Marukesu:
fixes #164


marbetschar (approved):
Neat, thank you very much! Works for me. From what I know @danrabbit uses GTasks extensively, so he might want to test this as well before we merge.

Issue #149, opened 2020-10-09
Scheduled List: Fix for S-Expression


marbetschar:
As [EDS added support for `has-end?`, `has-due?` and `has-duration?` in `3.39.1+`](https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/254), we should also fix our s-expression for the Scheduled List in [`MainWindow.vala`](https://github.com/elementary/tasks/blob/89f48c92a28af71d8b6c460483fc2daec9271a51/src/MainWindow.vala#L167):

```
var query = "AND (NOT is-completed?) (OR (has-start?) (has-alarms?))";
```

According to [RFC554, either dtstart or due](https://tools.ietf.org/html/rfc5545#section-3.6.2) can be set - so our s-expression should be changed to:

```
var query = "AND (NOT is-completed?) (OR (has-start?) (has-due?))";
```

@danrabbit will Evolution Data Server `3.39.1+` eventually land in Odin?

marbetschar:
Sidenote: Whether EDS lands in Odin or not, in every case we want to remove `has-alarms?` from the s-expression - https://github.com/elementary/tasks/pull/147 does take care of this.

danrabbit:
@marbetschar at this point we're past the freeze in Ubuntu, so we'd have to backport it ourselves. It looks like we're currently on 3.36.4. My assumption would be that we wouldn't normally backport something like EDS because Canonical maintains security patches for what's in their repos, but @tintou would have more of an opinion on this

marbetschar:
iOS Reminders sets both properties (start and due), so I guess we are fine if we do the same. Then we can simply query for `(has-start?)`. The needed changes for this to work is implemented in #147.

Although I don't know about other Task/Reminder Apps. I guess the most interesting part here would be to know how a broadly used one on Android behaves (I guess thats [Tasks](https://tasks.org/)) to figure out if we are on the right track here.

tintou:
I would indeed avoid as much backport as we can as it would mean that we have to support it for 3 years ourselves…
We will be flatpak-native at some point which would simplify issues like this a bit.

PR #204, opened 2021-03-16
GTasks: only show due date

### LINKED ISSUE #199: GTasks: scheduling time always at midnight
## What Happened

Trying to create or change a scheduled task from GTasks results with the time set to 00:00 from the scheduled day. On the 
GTasks app it don't have a time setted.

## Expected Behavior

That the task save the time defined.

## Steps to Reproduce

1. Create a new task on a GTasks task list.
2. Set a scheduling day and time. (Tomorrow, 10:00, for example)
3. see the task save only the day. (Tomorrow, 00:00)

## Platform Information

ElementaryOS 6.0 Early Access. Tested with flatpak and non-sandboxed versions

Marukesu:
Fix #199


marbetschar:
@Marukesu if its GTasks which discards the time, isn't this supposed to be a bug in the GTasks API? Or is the time discarded at the server end?

Marukesu:
It's on server side, from [Google Tasks API reference](https://developers.google.com/tasks/reference/rest/v1/tasks):

>The Due date only records date information; the time portion of the timestamp is discarted when setting the due date. It isn't possible to read or write the time that a task is due via the API.

src/Widgets/TaskRow.vala
```diff
@@ -113,20 +127,27 @@ public class Tasks.TaskRow : Gtk.ListBoxRow {
...
-            var format = h24_settings.get_string ("clock-format");
-
-            if (is_scheduled_view) {
-                return _("%s").printf (
-                    value.format (Granite.DateTime.get_default_time_format (format.contains ("12h")))
-                );
-
+            if (is_gtask) {
+                if (is_scheduled_view) {
+                    return null;
```
marbetschar, line 132: Not 100% sure if it is related to this, but I can't see any GTasks in my Scheduled view. Is this supposed to work already?

marbetschar (changes requested):

src/Widgets/TaskRow.vala
Marukesu: Since we can't check for due on the s-expression (#149), they don't appear even on master, I only added this for completeness.

marbetschar: Agree. That's why we set start date to the same value as the due date (https://github.com/elementary/tasks/pull/147). Isn't this possible using GTasks?

Marukesu: Since GTasks tasks are GData.TasksTask instead of Ical.Component, EDS needs to convert them, and only sets the due date on the conversion. So they do not have a start time when they come from the server

marbetschar: Alright, thanks for the clarification! I made an issue to keep track of this: https://github.com/elementary/tasks/issues/205

marbetschar (approved):

PR #206, opened 2021-03-19
GTasks: delete tasks list

Marukesu:


src/TaskModel.vala
```diff
@@ -239,6 +239,32 @@ public class Tasks.TaskModel : Object {
...
+
+                GData.TasksTasklist? tasklist = null;
+
+                unowned GLib.List<GData.Entry> gtasklists = service.query_all_tasklists (null, null, null).get_entries ();
+                foreach (unowned GData.Entry gtasklist in gtasklists) {
+                    if ("gtasks::%s".printf (gtasklist.id) == resource_extension.identity) {
+                        tasklist = (GData.TasksTasklist) gtasklist;
+                        break;
+                    }
+                }
```
marbetschar, line 256: Ideally, shouldn't this be solved by loading the `GData.TasksTasklist` directly with something like the following?

```vala
GData.TasksTasklist? tasklist = (GData.TasksTasklist?) service.query_single_entry (
    GData.TasksService.get_primary_authorization_domain (),
    resource_extension.dup_identity (),
    null,
    typeof (GData.TasksTasklist),
    null
);
```

However, this seems to throw the following error. Is this a bug in libgdata which should be fixed?

```bash
libgdata:ERROR:../gdata/gdata-service.c:588:_gdata_service_build_message: assertion failed (soup_uri_get_scheme (_uri) == SOUP_URI_SCHEME_HTTPS): ("gtasks" == "https")
Bail out! libgdata:ERROR:../gdata/gdata-service.c:588:_gdata_service_build_message: assertion failed (soup_uri_get_scheme (_uri) == SOUP_URI_SCHEME_HTTPS): ("gtasks" == "https")
Aborted
```

marbetschar (changes requested):
Looks good. However, there should be a simpler way to delete a tasklist which avoids loading and looping all available lists.

src/TaskModel.vala
Marukesu: I did it first and ended on the same error, so i stick with the way we do to rename the tasklist.
But looking how libgdata use it internaly, i don't think it's a bug, we only need to pass the full method uri instead, so:
``` vala
GData.TasksTasklist tasklist = (GData.TasksTaskList) service.query_single_entry (
    GData.TasksService.get_primary_authorization_domain (),
    "https://www.googleapis.com/tasks/v1/users/@me/lists/%s".printf (resource_extension.identity.replace ("gtasks::", "")),
    null,
    typeof (GData.TasksTaskList),
    null
);
```

src/TaskModel.vala
```diff
@@ -239,6 +239,38 @@ public class Tasks.TaskModel : Object {
...
+                } catch (Error e) {
+                    if (e is GData.ServiceError.NOT_FOUND) {
+                        throw new Tasks.TaskModelError.BACKEND_ERROR (
+                            "Task list '%s' is no longer available in Google backend.",
+                            id
+                        );
+                    } else {
+                        throw e;
+                    }
+                }
```
marbetschar, line 269: Don't know about the exact details of the error message, but I guess we are fine with whatever error is thrown here. Handling just this one particular error case differently seems to complicate things a bit too much.

In case this is needed, we can still re-add in the future.

marbetschar (changes requested):
Indeed, libgdata really works with hardcoded URI's internally ... interesting :thinking: 

Well, then I guess the reworked approach is the way to go! Left just one small comment regarding error handling - then we are ready to merge :partying_face:

marbetschar (approved):
Looks good, thank you very much!!!

PR #211, opened 2021-03-24
GTasks: don't allow exclude the default task list

### LINKED ISSUE #208: GTasks: Deleting of "default list" not possible
## What Happened

Whenever I try to delete one particular task list from GTasks, it throws the following error - any other task list is deleteable without any problem:

Screenshot from 2021-03-22 20-54-21

## Expected Behavior

If the task list is not deleteable, the delete functionality should be disabled in the UI.

Marukesu:
fix #208


src/TaskModel.vala
```diff
@@ -303,6 +303,30 @@ public class Tasks.TaskModel : Object {
         return false;
     }
 
+    private bool is_default_task_list (E.Source task_list, E.SourceRegistry registry) {
```
marbetschar, line 306: Just some minor improvements:

If I got this straigt, this function is only useful for GTasks - so how about renaming it to `is_gtasks_default_task_list` and I think we can just throw the Error so its handled by the calling function.

So we end up with something like this:

```
private bool is_gtasks_default_task_list (E.Source task_list, E.SourceRegistry registry) throws Error
```

marbetschar (changes requested):

marbetschar (approved):
Thank you!

PR #212, opened 2021-03-24
GTasks: Use X-EVOLUTION-GTASKS-POSITION for ordering

Marukesu:
Allow to GTasks task list to use the same ordering of the server.
the only bug is with sub-tasks that we don't manage and are added as a non-sub-task so breaking the ordering a little


src/ListView.vala
```diff
@@ -280,18 +289,24 @@ public class Tasks.ListView : Gtk.Grid {
...
-
-            var apple_sortorder_b = Util.get_apple_sortorder_property_value (row_b.task);
-            if (apple_sortorder_b == null) {
-                apple_sortorder_b = Util.get_apple_sortorder_default_value (row_b.task).as_int ().to_string ();
-            }
+            if (is_gtasks) {
+                var gtask_position_a = Util.get_gtasks_position_property_value (row_a.task);
+                var gtask_position_b = Util.get_gtasks_position_property_value (row_b.task);
+
+                return gtask_position_a.collate (gtask_position_b);
```
marbetschar, line 296: How should we deal with `gtask_position_a` or `gtask_position_b` beeing `null` here?

marbetschar (changes requested):

src/ListView.vala
Marukesu: I don't think they can be null, if a task is created without setting a position, it's settled to be on the top of the tasklist, however, looks like evolution isn't updating the position for already known tasks

Marukesu: So i found the issue, the GTasks API doesn't count moving the task as a update, so EDS don't update the task if the only change is on position, we probably need to patch it on the system EDS rather the flatpak one because the backends are on the system rather than the sandbox, otherwise we need to get the position using libgdata directly.

marbetschar: If you not already did, can you open an issue upstream so it is fixed in a future version of EDS?

In terms of elementary, its unlikely we are going to backport EDS - therefore we need to go for the workaround, at least on Odin.

marbetschar: also, I'd prefer we make sure we are not running into potential `null` errors and set `gtask_position_{a,b}` to an arbitrary value in case it is really returned as null - just to be sure.

Marukesu: > If you not already did, can you open an issue upstream so it is fixed in a future version of EDS?

Upstream issue [opened here](https://gitlab.gnome.org/GNOME/evolution-data-server/-/issues/322).
> In terms of elementary, its unlikely we are going to backport EDS - therefore we need to go for the workaround, at least on Odin.

We probably want to do that with #200, and use them for automatic refresh instead of relying on the eds one, or at least trigger the position update on the same interval of the eds one.

> also, I'd prefer we make sure we are not running into potential `null` errors and set `gtask_position_{a,b}` to an arbitrary value in case it is really returned as null - just to be sure.

I updated the function to ever return a value, if we can't get the property, it's will return the default value of a task created without setting a position.

marbetschar:
@Marukesu thank you for all the work you put into this! I think we are really close to get this merged! During testing the current state of work I noticed two things:

**1st: Sorting is reversed in comparison to the WebDAV sort**

Screenshot from 2021-04-02 11-12-43

Probably not a big deal, cause not a lot of people will use different backends at the same time - but I'm wondering, if we can use the same sort direction or if there is anything in particular which prevents in doing so? Like breaking user's expectations on another device synced via GTasks?

**2nd: New tasks "jump around"**

When I add a new task, the initial sorting is wrong - causing the task to "jump around". Any chance we can fix this?

Screen record from 2021-04-02 11 28 00

Marukesu:
> I'm wondering, if we can use the same sort direction or if there is anything in particular which prevents in doing so? Like breaking user's expectations on another device synced via GTasks?
> 

The default position for a task created without position like we do is 0, so it will be on top, that's what happens on GTasks Android app too.

If we want to make it appears on bottom in all backends, we need to set the position to the last (That will resolve 2 too, since the last position will be a new position), if we want the inverse, so the change is on the WebDAV task creation.
 
> **2nd: New tasks "jump around"**

The problem here is, we can't update the tasklist before the call for `create_object`. And after the call, `object_added` was already emitted. Updating the positions on the signal callback requires a sync version of the function, and that freezes the UI.

A solution can be create a `Tasks.Tasklist` class (or abstract class with specialised derivatives) and manage the task update before we call the ui to update.

Marukesu:
now when 2 tasks have the same positon, it set the more recently modified to first. it fix the created task jumping, but the others tasks can jump if they position was modified before on gtasks.

marbetschar:
FYI I created a discussion regarding sort order - so we can use whatever is better suited for now: https://github.com/elementary/tasks/discussions/220

Marukesu:
This PR now only manage to get and sort the tasks positions. The update of the tasklist position will be done in another PR.

src/TaskModel.vala
```diff
@@ -66,6 +65,28 @@ public class Tasks.TaskModel : Object {
...
+                            update_gtasks_tasklist.begin (task_list, client, registry);
+                            return Source.CONTINUE;
+                        });
+
+                        update_gtasks_tasklist.begin (task_list, client, registry);
+                    }
+                } catch (Error e) {
+                    warning ("cant get registry: %s", e.message);
+                }
+            });
```
marbetschar, line 88: Not sure what happened, didn't we agree on solving refresh in another PR?

src/TaskModel.vala
```diff
@@ -559,6 +585,10 @@ public class Tasks.TaskModel : Object {
 
             client.generate_instances_for_object_sync (comp, start.as_timet (), end.as_timet (), null, recur_instance_callback);
         }
+
+        if (ECal.util_component_has_x_property (comp, "X-EVOLUTION-GTASKS-POSITION")) {
+            yield update_gtasks_tasklist (list, client, get_registry_sync ());
+        }
```
marbetschar, line 591: Why do we need to trigger a refresh here? Once a task is completed, isn't it enough to just hide it from the UI?

src/TaskModel.vala
```diff
@@ -567,6 +597,10 @@ public class Tasks.TaskModel : Object {
 
         debug (@"Updating task '$(comp.get_uid())' [mod_type=$(mod_type)]");
         yield client.modify_object (comp, mod_type, ECal.OperationFlags.NONE, null);
+
+        if (ECal.util_component_has_x_property (comp, "X-EVOLUTION-GTASKS-POSITION")) {
+            yield update_gtasks_tasklist (list, client, get_registry_sync ());
+        }
```
marbetschar, line 603: This should be triggered by EDS. From what I understand EDS does trigger the refresh automatically EXCEPT if the `X-EVOLUTION-GTASKS-POSITION` changes on the remote, right?

src/TaskModel.vala
```diff
@@ -578,6 +612,10 @@ public class Tasks.TaskModel : Object {
 
         debug (@"Removing task '$uid'");
         yield client.remove_object (uid, rid, mod_type, ECal.OperationFlags.NONE, null);
+
+        if (ECal.util_component_has_x_property (comp, "X-EVOLUTION-GTASKS-POSITION")) {
+            yield update_gtasks_tasklist (list, client, get_registry_sync ());
+        }
```
marbetschar, line 618: Same as for `update_Task`

src/TaskModel.vala
```diff
@@ -640,6 +678,62 @@ public class Tasks.TaskModel : Object {
         }
     }
 
+    private async void update_gtasks_tasklist (E.Source task_list, ECal.Client client, E.SourceRegistry registry) {
```
marbetschar, line 681: Since this method completely bypasses EDS and seems not to be related to sorting at all, we might want to solve the refresh problem in another PR which also takes other backends into account. EDS' refresh for a WebDAV backend does not work 100% as expected too.

src/Util.vala
```diff
@@ -192,6 +192,38 @@ namespace Tasks.Util {
...
+        return "00000000000000000000";
+    }
+
+    public bool update_gtasks_position (ECal.Component ecalcomponent, string position) {
+        var icomp = ecalcomponent.get_icalcomponent ();
+        if (icomp != null) {
+            var icomp_position = ECal.util_component_dup_x_property (icomp, "X-EVOLUTION-GTASKS-POSITION");
+
+            if (icomp_position != position) {
+                ECal.util_component_set_x_property (icomp, "X-EVOLUTION-GTASKS-POSITION", position);
```
marbetschar, line 219: Nice catch! Seems like we can simplify some of the Task's code using this neat utility function.

marbetschar (changes requested):
I'm not 100% sure why, but the GTasks refresh code pieces are still in place. Maybe Git somehow mixed together my branch with the one you force pushed. In general its never a good idea to do a force push, because it'll will have unintended side-effects for others.

When in doubt, simply make a new commit - we are [squashing commits](https://devtut.github.io/git/squashing.html) when we merge anyways.

src/TaskModel.vala
Marukesu: Yeah, I think the merge take the commit back.

marbetschar (approved):
Looks good - thanks for bearing with me!

PR #217, opened 2021-04-02
Sort Tasks using Drag & Drop

marbetschar:
This PR adds drag and drop sorting for tasks within the same list. The basic logic is in place and works. But the design could use some more love to make it work better: Right now its not entirely clear where the drag and drop areas exactly are.


marbetschar:
This could be helpfull to highlight the drop area between two tasks:
https://blog.gtk.org/2017/06/01/drag-and-drop-in-lists-revisited/

marbetschar:
I'm closing this PR because it will need a major rework for Gtk4 anyway.

Issue #240, opened 2021-05-19
Error message when a task is unchecked


happycod3r:
## What Happened
When in task list view, if the users check a task that is already done he will receive an error.
error message:
![Screenshot from 2021-05-19 22-14-20](https://user-images.githubusercontent.com/20876378/118892279-2eb1d400-b8f0-11eb-8a97-f6e8971791cf.png)

## Expected Behavior
I was expecting the task to just unchecked without any error.

## Steps to Reproduce

1. Toggle "Show Completed".
2. Check a task as done.
3. Click the same checkbox again to uncheck the task.
4. An error message will appear.

## Platform Information
elementary OS 6 Early Access
Using manually cloned and compiled version

PR #241, opened 2021-05-19
fixing uncheck error message

happycod3r:
fixing issue [240](https://github.com/elementary/tasks/issues/240)

### After

taskdoneundone

### DIFF
--- a/src/TaskModel.vala
+++ b/src/TaskModel.vala
@@ -507,7 +507,7 @@ public class Tasks.TaskModel : Object {
 
             task.set_completed (new ICal.Time.null_time ());
 
-            yield client.modify_object (comp, ECal.ObjModType.ONLY_THIS, ECal.OperationFlags.NONE, null);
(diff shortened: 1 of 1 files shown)


marbetschar:
@happycod3r thanks for digging into the code and sending a PR! Can you please elaborate which backend you are using when receiving the `Cannot modify calendar object: Not supported` error? Are those tasks stored locally?

For context: I never experienced this error and I'm not sure how this change reflects repeating tasks of the various backends. Therefore I like to do some further testings to make extra sure we are not running into a regression of some sort with the change.

marbetschar:
I tested your change for Local, Google and CalDAV backends and they all behave the same like current master. So I'm merging this in - thanks again for the PR!

marbetschar (approved):

happycod3r:
Hey @marbetschar,
Sorry for the late reply, I've tested it locally. 
Glad to help :D

PR #245, opened 2021-05-28
Disable horizontal scrolling in task list

### LINKED ISSUE #231: Note that exceeds the length of the window truncates note
## What Happened
I entered a note that had enough text such that the entirety of the note could not be displayed on one line.  The string gets truncated to show the end of the note.

## Expected Behavior
I think that either the entire note should be displayed, an ellipsis should be displayed, a tooltip should display the entirety of the note in a tooltip (with formatting applied), or omit the display of the note, showing a note icon instead.

**Task Edit:**

image

**Task After Edit:**

image

## Platform Information
* elementary 6.0 Beta
* Latest release

mcclurgm:
~Fixes #231~

This disables horizontal scrolling in the task list, so the window can't be smaller than the tasks. This stops the weird truncation behavior. I assume it's fair to say that we don't need horizontal scrolling in this list; if not, I'll look for another solution to that bug.

I don't think this is quite an ideal solution to that issue, since it means that the minimum window width is rather large. But it at least stops the buggy behavior.


jeremypw:
@mcclurgm Please could you add Before and After screenshots showing the improvement this PR makes? It does not seem to make a visible difference for me (the task description still truncates rather than ellipsizes).

jeremypw:
After some testing, it seems that the problem is the transition in the `description_label_revealer`.  If this is removed then the label ellipsizes properly.

mcclurgm:
Hmm. To be honest, I didn't check that effect. My focus is more on truncating the entire contents of the window, which this PR does fix. I think your issue may be separate.

Here's the difference. I shrunk both of these windows down to about the smallest they'll get. With scrolling (original):
with-scroll
Without scrolling (new, restricts the width to the width of the widgets):
no-scroll

jeremypw:
@mcclurgm Oh, I see - you seem to fixing a different issue to #231. That issue seems to be about long descriptions, exceeding the available width, not truncating/ellipsizing properly when not editing.  Your screenshots shows a short description so its not clear whether this PR addresses the problem.

jeremypw:
Full Description shown in Edit Mode (master)
TaskLongDescriptionEdit

Description label not properly aligned and ellipsized (master)
TaskLongDescriptionMaster

After removing revealer transition
TaskLongDescriptionNoTransition

mcclurgm:
Aha, yeah, it seems like I'm not solving the issue I thought I was. But since this solves a different issue, I think it can still be merged, just without closing the issue.

jeremypw:
OK, I'll approve this and submit a separate PR for the other issue.

jeremypw (approved):

PR #248, opened 2021-06-30
Tasks appdata language reworked

comradekingu:


danrabbit (approved):
Thanks for this improvement! Looks good

PR #251, opened 2021-07-08
ListSettingsPopover: Use Granite.SwitchModelButton

danrabbit:


jeremypw (approved):
Works as expected.  Nice simplification.

PR #256, opened 2021-07-20
Create release.yml

danrabbit:
I'm using both the elementary release action and the flatpak builder action here because, even though we don't really need the deb branch stuff anymore, the elementary release action also includes tagging a release in GitHub


btkostner (approved):
Should be good now

PR #260, opened 2021-07-28
Release 6.0.1

danrabbit:
It sounds like this is needed so that Tasks can be used in languages other than English


marbetschar:
Would be great to sneak #261 into the release ;)

cassidyjames (approved):

Issue #228, opened 2021-04-26
Scheduled List: Sorting does not work correctly with multiple backends


marbetschar:
## What Happened

If one uses multiple backends the Scheduled list shows multiple sections for the same date (e.g. "Today" appears multiple times).

## Expected Behavior

All tasks from all lists and backends should be displayed within one section per day for a better overview.

PR #261, opened 2021-07-28
Fix Scheduled List sort

marbetschar:
This PR fixes https://github.com/elementary/tasks/issues/228. Turned out we just need to make sure to sort using due date when on Scheduled List (that is: no specific source is set for the list view).

**Before**

Before

**After**

After


danrabbit (approved):
Can confirm the fix!

PR #279, opened 2021-08-21
Avoid accidentally deleting a list

### LINKED ISSUE #244: Ctrl+Backspace to delete list is confusing
## Problem

Currently, Ctrl+Backspace is bound to delete the current list. This still happens when I'm in a text field, overriding the usual "delete last word" behavior. I'm not sure if this counts as a bug that it's not superseded by the text field or a request to remove the behavior, but either way the current behavior causes a lot of problems for me. I regularly use that shortcut to delete the last word (I'm not a super accurate typist, so I do this all the time). Instead, it deletes the list I'm working on with no warning.

## Proposal

Either remove the shortcut altogether or disable it when a text field is selected.

## Prior Art

n/a

### LINKED ISSUE #270: Pressing the Delete key (not backspace) deletes the whole list and it's tasks.
## What Happened

I was creating Tasks and editing the description when I pressed the Delete (not backspace) key and the whole category was deleted including all the tasks, Ctrl Z did not bring them back. Reproduce by creating a task list, then create a task. While typing what the task is, press the delete button and the whole task list should be deleted.

## Expected Behavior

It should delete the letter to the right of my cursor.

## Steps to Reproduce

1. Open Tasks
2. Create a list (don't even have to name it)
3. Start typing out the task name
4. Press the Delete key (not backspace)

### Logs

If you can't reproduce (pretty easy to reproduce) I'll try and get logs.

## Platform Information

elementary OS 6
Tasks Latest Release

marbetschar:
Fixes #270, fixes #244 

This PR fixes https://github.com/elementary/tasks/issues/270 by introducing the following changes:

1. The `DELETE` key shortcut is completely removed - thus using it while editing text does no longer accidentally delete a list
2. Added a confirmation dialog when deleting a list
3. Made the `<Control>Backspace` shortcut discoverable in the list settings popover


Merazi:
I might be a little bit late to the party but...

> @mcclurgm just seen your issue #244:
> 
> Since this PR attempts to fix issues caused by accidentally using the delete shortcut for a list - which shortcut would you expect? I went for Ctrl + Backspace because both keys are available on all keyboard, but I'm open for better suggestions.

About the suggestions... What about `Ctrl + D`? D being **D** -elete.
_AFAIK_ that key combination is not being used for anything else in the app, so it may work... What do you think?

mcclurgm:
👍🏻 that makes a lot of sense to me. Perhaps we could overload it with both? Or either way. I'm not aware of any `ctrl+d` shortcuts in the standard Gtk suite. It is a little bit un-traditional, but I do feel like I've seen it used in some apps before, I just can't remember where.

src/MainWindow.vala
```diff
@@ -318,18 +318,34 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
-                    show_error_dialog (
-                        _("Deleting the task list failed"),
-                        _("The task list registry may be unavailable or unable to be written to."),
-                        e
-                    );
-                }
-            });
+            var message_dialog = new Granite.MessageDialog.with_image_from_icon_name (
+                _("Really Delete List?"),
+                _("The list '%s' will be completely deleted, including all its tasks.").printf (source.display_name),
```
mcclurgm, line 323: Some wording changes to the new dialog. This is just a suggestion, but I think it better follows the HIG. The standard capitalization is sentence case for the primary message, and I think that placing the list name there instead of the smaller text is clearer about the resulting action (although that is clearly up for debate).


```suggestion
                _("Permanently delete list \"%s\"?").printf (source.display_name),
                _("The list and all its tasks will be permanently deleted."),
```

mcclurgm:
This makes sense to me for the most part. I like that it's disabled when a task is selected at all, that makes a lot of sense to me. The only remaining case I can think of that will unintentionally trigger the dialog is when you're editing the title of a list, which granted won't happen much.

src/Widgets/ListSettingsPopover.vala
```diff
@@ -115,11 +115,16 @@ public class Tasks.Widgets.ListSettingsPopover : Gtk.Popover {
             margin_top = 3
         };
 
-        var delete_button = new Gtk.ModelButton () {
-            action_name = MainWindow.ACTION_PREFIX + MainWindow.ACTION_DELETE_SELECTED_LIST,
-            text = _("Delete List")
-        };
-        delete_button.get_style_context ().add_class (Gtk.STYLE_CLASS_DESTRUCTIVE_ACTION);
+        var delete_list_accel_label = new Granite.AccelLabel.from_action_name (
+            _("Delete List"),
```
mcclurgm, line 119: I think that since this now sends a dialog, it should have an ellipsis.

```suggestion
            _("Delete List…"),
```

src/MainWindow.vala
mcclurgm: On second thoughts, maybe "permanently" is only necessary in the subtitle.

marbetschar:
@mcclurgm found an easy way to also deactivate the shortcut while the list title is edited - so this is fixed now too.

@Merazi I did a little research and ended up preferring the `Ctrl + Backspace` variant because we also use `Backspace` for delete in other elementary Apps (e.g. to delete a message in elementary Mail). So I'd rather not introduce a totally unrelated shortcut here.

Merazi:
> @mcclurgm found an easy way to also deactivate the shortcut while the list title is edited - so this is fixed now too.
> 
> @Merazi I did a little research and ended up preferring the `Ctrl + Backspace` variant because we also use `Backspace` for delete in other elementary Apps (e.g. to delete a message in elementary Mail). So I'd rather not introduce a totally unrelated shortcut here.

Oh ok, understandable, the final solution is wonderful :smile: thanks for taking the time.

mcclurgm:
Sorry for editing your description, I was trying to get it to automatically close the issues. I can't tell if it worked...

Anway, this checks out for me wise. The only concern I have (which I just realized) is, is it possible to delete an entry with a key binding? I would almost expect that to be ctrl+backspace, and a list to be something else. But perhaps we can deal with that separately from this PR.

marbetschar:
There isn't any shortcut to delete a task. If there would be one, I'd be tempted to use only Backspace - since that's the shortcut in Mail too.

But yeah, that's one for another PR ;)

mcclurgm (approved):
Alright this seems good to me. I'm not familiar with Gtk's keybinding system so I can't really review the code, but it appears to work as advertised 👍🏻

marbetschar:
I'm merging this in for now - we can always iterate on subsequent PR's :)

PR #281, opened 2021-08-26
[Testing] HyperTextView

marbetschar:
Added the current implementation of HyperTextView for testing purposes. See https://github.com/elementary/granite/pull/507 for details.


mcclurgm:
If this helps track down the tooltip bug, on my system there seems to be an issue with position offset, but only with the hover detector. It starts recognizing text halfway down the line, and continues recognizing the link halfway down the next line. This means that if I position the cursor over what seems like the middle of the text, odds are that it won't see the link. The ctrl+click operation works fine, presumably because you get the iterator a different way.

This seems to be unique to Tasks. There may be some way that we're generating an offset that Gtk isn't aware of or treating properly.

Here's an example:
Screenshot from 2021-09-19 11 49 41

jeremypw:
Definitely a problem getting window coords.  This is what happens if you add enough text to cause scrolling:
TasksHyperTextView
The first few lines containing hypertext are not recognised as such.

I'll have a look into this and try and fix it.

jeremypw:
I have pushed a commit that seems to fix the tooltip problem.

marbetschar:
@jeremypw thank you very much! Tested your fix and it works smoothly - it is now also part of the official HyperTextView PR over at the granite repo :partying_face: https://github.com/elementary/granite/pull/507

marbetschar:
Supserseded by https://github.com/elementary/tasks/pull/314

PR #287, opened 2021-08-29
Fix account header in sidebar

marbetschar:
There was an issue where the same account name was displayed multiple times in the sidebar after a new task list was added. This only happens if the headers in the sidebar were already displayed and the user adds a new list with a name which occurs newly as the first item in the list. In this case we need to make sure we remove any previously added header label - and that's exactly what this PR does:

**Without the fix:**

https://user-images.githubusercontent.com/392542/131247492-8509ecac-f7dc-4831-add9-7f5b984e7119.mp4

**With the fix:**

https://user-images.githubusercontent.com/392542/131247495-f719c353-1d2c-46d7-8843-d7fecdd720a6.mp4


JoseExposito:
LGTM, thanks!

JoseExposito (approved):

Issue #201, opened 2021-03-09
Refactor: Move ListView to Widgets


marbetschar:
Since PR https://github.com/elementary/tasks/pull/197 `ListView` is no longer just a view that gets updated, but a widget which has many instances. Therefore it should be moved to `Tasks.Widgets`

I imagine the scheduled view could be its own widget, that might take out a lot of conditional statements.

PR #289, opened 2021-08-31
Refactor: Splits TaskListGrid into two classes

marbetschar:
This PR fixes tasks 2.1 and 2.2 of https://github.com/elementary/tasks/issues/201. There is some more refactoring which can be done here for the sake of code quality. Will do so in another PR to make this easier to review.


JoseExposito (approved):
I'm not very familiar with the code base, but this is moving code around, so I'm confident merging it.
Thanks for splitting the changes in 2 commits, way easier to review :rocket:

PR #293, opened 2021-09-06
Fix cursor placing on focus

marbetschar:
During my HyperTextView testings I noticed the cursor placing in the description text view does not behave as expected on focus. This PR fixes this.

**Without this PR**

https://user-images.githubusercontent.com/392542/132178799-8fba8b44-2a44-4cc9-83ac-f36bcaafb309.mp4

**With this PR**

https://user-images.githubusercontent.com/392542/132178429-2496e6bc-d0a6-4133-b126-96c9e6c39691.mp4


danrabbit (approved):
Works as expected here. Seems like this was originally a workaround for an issue that was solved by making the row itself not be allowed to steal focus

Issue #294, opened 2021-09-08
Crash: Creating pipes for GWakeup


Letterus:
## What Happened

Tasks crashes on startup. The list of calendars is shown correctly, but the tasks most times are not shown. It seems Tasks crashes while loading them.

I tried to set the inode number very high, but `df -i` always shows way enough free inodes, so I don't really know if this is the problem. 

## Expected Behavior

I'd expect my tasks to appear and the UI to react to my input.

## Steps to Reproduce

1. Add a Nextcloud account with really a lot of calendars and tasks (some thousands).
2. Start Tasks
3. Wait for the UI to load the tasks, but it crashes before it finishes.

### Logs
```
~$ io.elementary.tasks 

(io.elementary.tasks:16885): GLib-ERROR **: 22:31:23.069: Creating pipes for GWakeup: Too many open files
Trace/Breakpoint triggered.
```

## Platform Information

elementary OS 6 Odin, all updates installed.

marbetschar:
> Add a Nextcloud account with really a lot of calendars and tasks (some thousands).

That's a lot of tasks - you must be really busy! 😄 To track this down, the following information would be usefull:

- How many task lists are there?
- How many tasks per list are there (approx)?
- How many of those tasks per list are completed (rough percentage)?

In addition, can you post the output of the following commands (of course without sensitive information if there's any)?

```bash
export G_MESSAGES_DEBUG=all
io.elementary.tasks
```

Letterus:
Thank you for your reply. That debug log is helpful. I won't upload it here as it contains too much sensitive information, but it appears that Tasks already crashes on trying to load the first task list/calendar (my task inbox "Eingang") which contains 6 active and 1165 completed tasks.

It seems to me the main problem is that Tasks tries to apply those watches even to the completed tasks:

```
(io.elementary.tasks:5161): dconf-DEBUG: 14:23:14.218: watch_fast: "/org/gnome/desktop/interface/" (establishing: 0, active: 0)
(io.elementary.tasks:5161): dconf-DEBUG: 14:23:14.218: unwatch_fast: "/org/gnome/desktop/interface/" (active: 0, establishing: 1)
(io.elementary.tasks:5161): dconf-DEBUG: 14:23:14.219: watch_established: "/org/gnome/desktop/interface/" (establishing: 0)
(io.elementary.tasks:5161): dconf-DEBUG: 14:23:14.220: watch_fast: "/org/gnome/desktop/interface/" (establishing: 0, active: 0)
[...]

PR #295, opened 2021-09-14
Load only non-completed tasks by default

marbetschar:
Fixes a crash experienced by users with a lot of tasks per list (https://github.com/elementary/tasks/issues/294). The root cause there is, that we are loading all tasks per default - even the completed ones.

This PR mitigates the situation by only loading the non-completed tasks by default. If the user enables "Show Completed", then a new query is executed which also retrieves the completed tasks.

Please note, loading the completed tasks as well would still lead to the crash. But in Gtk3 there is no builtin way to mitigate this, because Gtk.ListBox simply renders whatever we throw at it - instead of only doing so for the visible area only. A more elegant fix would be to use [Gtk4's ListView](https://valadoc.org/gtk4/Gtk.ListView.html) - but at the time of writing this is not possible, since [we can't pack Tasks as Flatpak](https://github.com/elementary/switchboard-plug-onlineaccounts/issues/209) right now.

Therefore this strategy at least [solves the immediate problem](https://github.com/elementary/tasks/issues/294#issuecomment-918549321) and we can refactor this later on.

### DIFF
--- a/data/tasks.appdata.xml.in
+++ b/data/tasks.appdata.xml.in
@@ -22,6 +22,7 @@
           <li>Fix an issue where an account name header could be displayed multiple times in the sidebar</li>
           <li>Fix an issue with long descriptions being truncated at the beginning instead of the end</li>
           <li>Fix an issue where removing descriptions wasn't saved</li>
+          <li>Improve performance in lists with lots of completed tasks</li>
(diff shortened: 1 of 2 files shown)


danrabbit (approved):
Makes sense and works for me! Nice fix

Issue #168, opened 2020-11-25
Drag & Drop Task to another Task List


marbetschar:
It should be possible to move a task to another list by using Drag & Drop.

alainm23:
Is there a property in `ECal.Component` or `ICal.Component` to store the order of the task in the list?

marbetschar:
@alainm23 I don't know if there is a standard way of sorting. I can't find anything sort-related in the [RFC](https://tools.ietf.org/html/rfc5545#section-3.6.2). And iOS/Apple uses a custom property called `X-APPLE-SORT-ORDER`.

You can access [custom properties similar to this](https://github.com/elementary/tasks/blob/f3817889b4c78e7cfc9f3a380668ef5f2ccb2c05/src/Util.vala#L156-L195).

alainm23:
@marbetschar For this to work we have to change how the task editing works, currently the [grab_focus](https://github.com/elementary/tasks/blob/master/src/Widgets/TaskRow.vala#L248) event is used to enable the editing of a task. This does not allow a task to drag naturally. In Planner I use the [row_activated](https://github.com/alainm23/planner/blob/master/src/Views/Inbox.vala#L331) event to enable editing of a task, we could make that change first and D&D could work.

marbetschar:
@alainm23 sounds good to me. The current task edit abort handling seems to interfere with row activation anyways.

BTW: I've added some utility functions to access X-Properties in #174 (e.g.: Util.get_icalcomponent_x_property) - so once this is merged it should also be easier to access `X-APPLE-SORT-ORDER` or similar props.

PR #296, opened 2021-09-15
Drag Task to another List

marbetschar:
This PR fixes https://github.com/elementary/tasks/issues/168. The code itself works even if the task is draged onto a list which is stored at another backend. Also its intended to move multiple tasks at a time, in case we want to allow multi select in the future.

~~However, UX-wise there is still room for improvement. Since I don't have a lot of experience with drag and drop in GTK I adhered to http://www.cs.hunter.cuny.edu/~sweiss/course_materials/csci493.70/lecture_notes/GTK_dragndrop.pdf which explains the technical mechanisms at play quite well - but...~~

~~... I could use some UX/Design help for things like:~~

- [x] ~~Highlight the drop area~~
- [x] ~~Generate a preview of the dragged task~~

I managed to get the design stuff done (see result in video) - so I think we are ready for review here:

https://user-images.githubusercontent.com/392542/133591767-5282a876-43fa-42c3-8984-98f93ae69b42.mp4


danrabbit (approved):
Nice work!

PR #298, opened 2021-09-21
Notifications

marbetschar:
This PR is add notifications for tasks which are due:

Screenshot from 2021-09-21 23-45-39


mcclurgm:
I haven't made it to that yet, but I did notice a few things that are getting in the way.

Now each time I click the close button I get a warning:
```
(io.elementary.tasks:20557): GLib-GObject-WARNING **: 07:53:59.574: ../../../gobject/gsignal.c:2736: instance '0x555d39a311a0' has no handler with id '1200'
```
The id number varies.

Also, I'm getting an assertion failure. When I try to save with a time set, one of the assertions in my `icaltime_get_timezone` function fails and crashes the app. Specifically, the `ICal.Time` has an empty TZID, but its timezone is not null. I didn't account for that in the original code. I'm not sure how that can happen, do you have any suggestions? Or could you point me to where we generate timezones for tasks so I can see what's being set?

marbetschar:
@mcclurgm I guess the time issue with the notification is related to https://github.com/elementary/tasks/issues/312?

mcclurgm:
I hadn't made that connection--you're probably right.

danrabbit (changes requested):
It looks like this needs merge conflicts resolved after the timezone stuff was merged :)

marbetschar:
This is now ready for review :partying_face: 

- Fixed Merge Conflicts
- Thanks to the fix of @mcclurgm notifications are now fired at the right time
- I also added a fix which prevents firing the same notification again when Tasks is started in the foreground

Sidenote: The aforementioned fix does not prevent the same notification to be fired in all cases, because I only store sent notifications in memory. Therefore we'll see the same notification for due tasks within the last 12 hours whenever `io.elementary.tasks --background` is restarted. But I think that's not a blocker for now and needs some more real world testing if this is really an issue.

danrabbit (approved):
Since we just got the release out, let's get this in and start testing :)

PR #300, opened 2021-09-22
MainWindow: copy and icon for delete confirmation dialog

danrabbit:
Port some of the stuff from #15

* Use `edit-delete` icon with a question badge
* Include language that warns about the implication on others
* Use "Anyways" like we do in other destructive action confirmations


src/MainWindow.vala
```diff
@@ -319,15 +319,17 @@ public class Tasks.MainWindow : Hdy.ApplicationWindow {
...
+                _("The list and all its tasks will be permanently deleted. If you've shared this list, other people will no longer have access."),
+                "edit-delete",
                 Gtk.ButtonsType.CANCEL
             ) {
+                badge_icon = new ThemedIcon ("dialog-question"),
                 transient_for = this
             };
 
-            unowned Gtk.Widget trash_button = message_dialog.add_button (_("Delete List"), Gtk.ResponseType.YES);
+            unowned Gtk.Widget trash_button = message_dialog.add_button (_("Delete Anyways"), Gtk.ResponseType.YES);
```
cassidyjames, line 330: Looks like we use "Anyway" elsewhere. And now after looking at it repeatedly, they both look wrong to me but at least this is consistent lmao.

```suggestion
            unowned Gtk.Widget trash_button = message_dialog.add_button (_("Delete Anyway"), Gtk.ResponseType.YES);
```

cassidyjames (changes requested):
Tiny nit but otherwise good

cassidyjames (approved):

PR #311, opened 2021-11-06
Free libical cache before exiting the application

tintou:
This allows to not technically leak memory.


src/Application.vala
```diff
@@ -117,6 +117,8 @@ public class Tasks.Application : Gtk.Application {
     public static int main (string[] args) {
         GtkClutter.init (ref args);
         var app = new Application ();
-        return app.run (args);
+        int res = app.run (args);
+        ICal.Object.free_global_objects ();
```
jeremypw, line 121: Presumably this will not interfere with other apps using ICal?

tintou: It shouldn't, they're not sharing the memory

jeremypw (approved):

PR #313, opened 2021-11-27
Use floating timezones when saving timed tasks

### LINKED ISSUE #312: Improve time zone handling: floating by default?
### Problem

Tasks currently has an ill-defined way of handling timezones. (See elementary/wingpanel-indicator-datetime#252). It assigns reminders a timezone of UTC, but treats that as local. We should define the way it works.

Also, it includes timezones in reminders by default. So if you move between time zones, your reminders will change. I think that in most cases that you want a reminder for a task, it's the time of day that matters, not the time on a universal clock. 
For example, taking trash out, cooking, taking medicine, or other such tasks are all based on the time of day where you are. As an extreme example, if I move across the world, I wouldn't expect to start cooking dinner at 6 am instead of 6 pm. This reminder makes more sense as a floating timezone. 
Home examples are easier to say that they should be based on time of day not universal time. Work tasks tasks are less clear.

### Proposal

The most basic thing we could do is assign tasks the local timezone. This would fix bugs with treating UTC as local and not converting.

I would suggest that we leave task reminders with a floating timezone. This would fix the issues with changing reminder time when changing timezones.

However, making all tasks floating may have some drawbacks. There will always be cases where people will want to set a reminder at a certain universal time and have the time of day change. Also, if you share a task list with someone else in a different timezone, you'll get reminders at different times. The question then becomes which case do we think is more likely: universal or time of day?

We could implement a way of selecting a timezone (see Todoist's method for a simplified version of this), but I think that may be more detailed than we need. I don't know how to go about researching this, but I would expect that a floating time will work for most use cases.

I would appreciate design input about this. Implementing the change should be relatively easy, since our backend already supports all the timezone behaviors I'm describing, so the change would be restricted to the UI.

### Prior Art (Optional)

Other tasks apps I use are Things and Todoist. Both use floating times by default (although you can change that in Todoist, see [help article](https://todoist.com/help/articles/set-a-fixed-time-zone-or-floating-time-for-a-task)).

mcclurgm:
Fixes #312 

This PR currently does the bare minimum to fix this issue. It uses a floating timezone when saving every task it gets. I copied the updated `datetimes_to_icaltime` function from elementary/calendar#653 to allow for floating timezones. I intend to iterate it with feedback.

A few action items for review before I would consider this complete:

1. Overwriting all timezones may not be desirable. Do we want to overwrite every event with floating, or should we preserve existing timezones? (I believe we should preserve the data that's there, but haven't implemented it currently.) 
2. If we don't overwrite timezones, then we will probably need to do something to manage existing tasks that are erroneously in UTC, since Tasks won't allow users to fix that. I would appreciate guidance from @elementary/ux on this.

Still to do for myself:

- [x] Update the task row to convert task times to local **(I assume this is the desired behavior?)**

If you know of any other places in Tasks where times are handled, let me know and I will investigate for required changes. I'm not super familiar with the Tasks codebase, so I'm not sure where to look beyond what I have changed here.


marbetschar:
@mcclurgm does it work if we ask the user to remove the time, save the task and then re-add the time in case they run into this?

I doubt there are that many tasks affected that it s reasonable to add an automated solution :shrug:

mcclurgm:
I'm not sure what the solution would be. Adding an automated solution would be a bunch of work for probably little benefit. Adding a warning like that could make sense. Also, since we now convert time zones to local, users will see that their events' times have changed. This could be a prompt for them to change them, but it also looks like the app is destroying their data (when really it's correcting a past bug).

I pushed an update that now converts to local times. I changed all the places I noticed where we display times, but please let me know if there are any I missed.

marbetschar:
@danrabbit what do you think is the best approach here? Doing an automatic conversion or doing nothing automatic at all and just somehow inform the user? (Either programmatically or just in a blog post for example)

mcclurgm:
For a status update: the only things remaining to do here before I consider it ready for merge review

- [ ] Preserve the timezone for a task, if it already exists. (I'm having trouble finding a place to store this, currently WIP)
- [ ] Data transition plan if necessary

marbetschar:
@mcclurgm can you please elaborate why we need to store the timezone _if it already exists_? Is this needed for newly added tasks as well - or just for tasks created before this fix?

mcclurgm:
Sorry that wasn't very clear at all. This storage is internal, it's just a matter of getting the relevant information from the component to where it needs to go. If one exists we need to pass it to `datetime_to_ical`, which needs it because it constructs a new `ICal.Time` from scratch every time. 

I suppose there's a really simple option for this that I just glossed over yesterday. Since the saving only happens in the `save_task` function, we could probably just poll the `task` property there and be done with it. That would be simplest. 
To prevent remaking objects from whole cloth every time we could also rearchitect the converter to accept an `ICal.Time` and modify it, which would inherently preserve all the relevant info from the component. I expect that's more work than it's worth, though.

One more warning that just occurred to me: once we start accepting and converting arbitrary timezones, we run into all sorts of conversion issues from libical to glib (see elementary/calendar#709). This is signing ourselves up for a fair bit of trouble, but since we don't control where people get their tasks from, I don't see how we can get around it.

mcclurgm:
I just pushed an update the does the simple version. It's not quite the prettiest, but I don't think there's much of a way to simplify it, since there are a few branching cases that we need to handle.

danrabbit:
Yeah I think we can just mention in a blog post to double check the timezone of your tasks. I don't think we need to spend a bunch of time on a programmatic solution here

danrabbit (approved):
This looks reasonable and events appear to have the right times in my testing. :rocket:

PR #316, opened 2021-12-08
Release 6.1.0

danrabbit:
https://github.com/elementary/tasks/compare/6.0.4...master

- [x] https://github.com/elementary/tasks/pull/296
- [x] https://github.com/elementary/tasks/pull/313


marbetschar:
Notifications would also be nice to have - if you have some time to review them as well :)

danrabbit:
I think because of how it sounds that we need to do some more design work and testing with notifications that it would be better to release first and then we can merge and test notifications to be ready for the next release cycle

cassidyjames (approved):

Issue #306, opened 2021-10-22
Provide refresh/reload button


Suzie97:
### Problem

Currently, Tasks does not seem to refresh tasks live when things are changed in the caldav backend. This makes it hard, as I need to open evolution and then do the refresh manually.

### Proposal

Provide a refresh button to reload/refresh tasks. Ideally, doing it live when things change in the backend would be neat, but for now, atleast providing a button to do looks important.

### Prior Art (Optional)

elementary OS Mail App
Apple Reminders reloads automatically when things change in the backend

metalune1:
Why hasn't this ever been worked on?

PR #330, opened 2022-04-28
Auto refresh task lists and collections

marbetschar:
This PR introduces an automatic way of keeping task lists current: Instead of just relaying on a regular sync interval we try to be a bit smarter and automatically synchronize a task list whenever it is selected by the user and a network connection is available - or if the network becomes available again.

This fixes https://github.com/elementary/tasks/issues/306 as it basically obsoletes an extra refresh button.

### DIFF
--- a/data/tasks.appdata.xml.in
+++ b/data/tasks.appdata.xml.in
@@ -13,6 +13,14 @@
     </p>
   </description>
   <releases>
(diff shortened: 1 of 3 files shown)


danrabbit (approved):
Works exactly as described. Tested with a Gmail account. Nice work!

PR #335, opened 2022-06-09
Square icon design

danirabbit:
Playing with the idea of a square icon where an element breaks out of the square


micahilbery (approved):
cute! :heart:

PR #356, opened 2023-04-10
Fix a small color issue

leolost2605:
Fix a small color issue that was noticeable using dark mode (see red arrow in the screenshot) by adding `Gtk.STYLE_CLASS_BACKGROUND` to `main_grid`.

Before:

seesdsadf

After:

image


danirabbit (approved):
makes sense, nice fix!

PR #360, opened 2023-06-10
Preparations for Gtk4

lenemter:
- Replace grids with boxes where possible
- Replace `margin` and `expand`  properties
- Update copyright header in touched files


danirabbit (approved):
:clap:

PR #361, opened 2023-06-12
Use flatpak portal for autostart and background permission

### LINKED ISSUE #357: Use the background portal
We are currently using the .deb package to install the background service, we should install it at first run via the background service portal https://flatpak.github.io/xdg-desktop-portal/#gdbus-org.freedesktop.portal.Background

leolost2605:
- Use the flatpak portal for installing the autostart file and requesting permission to run in background
- Always start the TodayTaskMonitor
- Uses libportal
- It's pretty much just copy/paste of elementary/mail#882 :)

We have to make sure though that if this gets merged elementary/portals gets a new release with/before tasks
Probably needs #362 

Fixes #357


danirabbit (approved):
Nice work!

PR #363, opened 2023-06-12
Application: use startup

danirabbit:


src/Application.vala
```diff
@@ -53,6 +53,21 @@ public class Tasks.Application : Gtk.Application {
...
         add_main_option_entries (OPTIONS);
+    }
+
+    protected override void startup () {
+        base.startup ();
+
+        Hdy.init ();
+
+        var granite_settings = Granite.Settings.get_default ();
+        var gtk_settings = Gtk.Settings.get_default ();
```
Marukesu, line 64:
```suggestion
        unowned var granite_settings = Granite.Settings.get_default ();
        unowned var gtk_settings = Gtk.Settings.get_default ();
```

src/Application.vala
```diff
@@ -53,6 +53,21 @@ public class Tasks.Application : Gtk.Application {
...
+
+        gtk_settings.gtk_application_prefer_dark_theme = granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK;
+
+        granite_settings.notify["prefers-color-scheme"].connect (() => {
+            gtk_settings.gtk_application_prefer_dark_theme = granite_settings.prefers_color_scheme == Granite.Settings.ColorScheme.DARK;
+        });
```
Marukesu, line 70: i think this is a good place to use the reduced enum name syntax.

```suggestion
        gtk_settings.gtk_application_prefer_dark_theme = granite_settings.prefers_color_scheme == DARK;

        granite_settings.notify["prefers-color-scheme"].connect ((obj) => {
            gtk_settings.gtk_application_prefer_dark_theme = ((Granite.Settings) obj).prefers_color_scheme == DARK;
        });
```

danirabbit: Ah yeah that's much nicer. Good call!

Marukesu (approved):
LGTM.

Issue #359, opened 2023-06-06
Duplicated tasks in Scheduled view when changing color of the task list


lenemter:
### What Happened?

Duplicated tasks in Scheduled view when changing color of the task list.

![image](https://github.com/elementary/tasks/assets/80143868/90dbbd1d-10eb-4cbe-a7f5-c95f12aa4f78)

### Steps to Reproduce

1. Open task list with tasks with due dates
2. Change its color
3. Open Scheduled view

### Expected Behavior

No duplicated tasks.

### OS Version

7.x (Horus)

### Software Version

Compiled from git

Claudio-code:
Analyzing the code it seems that this error happens because the ScheduledTaskListBox class has the signal methods task_list_added, task_list_modified and task_list_removed connected with your implementation and it shouldn't. TaskModel should be unique class to control state of tasks to avoid to connectors throw updated same event.

Claudio-code:
I managed to find Tasks.Widgets.TaskRow but it has the source with the hash and uid different from what comes by parameter in the function so the validation still wouldn't work

![image](https://github.com/elementary/tasks/assets/42524939/034fd7b6-3c18-4bfd-bdb5-c6d2f9915f9b)

Claudio-code:
if you remove the right TaskRow it stops duplicating anymore I can't find how to compare the TaskRow.source with the source that comes as a parameter

Claudio-code:
i think found solution if change get_children () to task_list.get_children () it method can remove repeated items

PR #364, opened 2023-06-22
Fixed Duplicated tasks in Scheduled

Claudio-code:
I tried fix it issue https://github.com/elementary/tasks/issues/359

For what i tested even removing this code all task updates are still reflected in the scheduled session.


lenemter (changes requested):
This is not how it's supposed to be done. All the code you removed is responsible for handling cases when new task lists are added or removed and it adds tracker of the list for them. Removing this code introduces some bugs, such as when user creates a new task list and adds tasks to it, these tasks will not appear in scheduled view.

The real issue here is in `remove_view` method. It doesn't properly destroy tasks. You can try to look into this :)

Claudio-code:
I understand I close the pr try another way thanks for the context

PR #370, opened 2023-07-31
Remove old autostart file

### LINKED ISSUE #369: Tasks automatically opens on system startup
### What Happened?

Tasks automatically opens now on system startup showing it's home screen when autostart is enabled.

### Steps to Reproduce

1. Make sure you have Tasks 63.2 and autostart is enabled (by default it is)
2. Restart your device

### Expected Behavior

It should start only in background, shouldn't open the UI

### OS Version

7.x (Horus)

### Software Version

Latest release (I have run all updates)

### Log Output

_No response_

### Hardware Info

_No response_

lenemter:
Fixes #369 

Tested locally and it actually removes the file.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -6,4 +6,6 @@ flatpak remove --system -y io.elementary.tasks//stable || true
 
 #DEBHELPER#
 
+rm /etc/xdg/autostart/io.elementary.tasks-daemon.desktop
+
 exit 0
--- a/debian/postinst
+++ b/debian/postinst
@@ -6,6 +6,6 @@ flatpak remove --system -y io.elementary.tasks//stable || true
 
 #DEBHELPER#
 
-rm /etc/xdg/autostart/io.elementary.tasks-daemon.desktop
+rm /etc/xdg/autostart/io.elementary.tasks-daemon.desktop || true
 
 exit 0


lenemter:
> What creates the autostart file?

The old autostart file (`/etc/xdg/autostart/io.elementary.tasks-daemon.desktop`) was created by meson. Since https://github.com/elementary/tasks/pull/361 the Background Portal creates the autostart file in `/home/lenemter/.config/autostart/` directory.

debian/postinst
```diff
@@ -6,4 +6,6 @@ flatpak remove --system -y io.elementary.tasks//stable || true
 
 #DEBHELPER#
 
+rm /etc/xdg/autostart/io.elementary.tasks-daemon.desktop
```
tintou, line 9: It still feels wrong but at least you need this
```suggestion
rm /etc/xdg/autostart/io.elementary.tasks-daemon.desktop || true
```

lenemter:
> It still feels wrong but at least you need this

@tintou Do you see a cleaner way? I had an idea to remove this file in meson but haven't tried it. Will it work?

danirabbit:
Just gonna merge for now so we're not release blocking and we can revisit later if there's a cleaner solution

PR #383, opened 2024-05-16
EditableLabel: Gtk4 Prep

danirabbit:
* Use Event Controllers
* We already have a good hover indicator for an editable item: the cursor. Remove the button revealer and use cursor
* Code style
* `grab_focus` isn't a signal in GTK 4


ryonakano (approved):
Code looks good and working as expected.

PR #386, opened 2024-06-06
EntryPopover/Generic: Gtk4 Prep

danirabbit:
* Use motion controller
* Construct button contents in GTK4 friendly way
* Code style
* use child property

### DIFF
--- a/src/Widgets/EntryPopover/Generic.vala
+++ b/src/Widgets/EntryPopover/Generic.vala
@@ -12,7 +12,6 @@ public abstract class Tasks.Widgets.EntryPopover.Generic<T> : Gtk.EventBox {
     public string placeholder { get; construct; }
     public T value { get; set; }
 
-    private Gtk.MenuButton popover_button;
     private T value_on_popover_show;
 
     private Gtk.EventControllerMotion motion_controller;
@@ -29,7 +28,7 @@ public abstract class Tasks.Widgets.EntryPopover.Generic<T> : Gtk.EventBox {
     }
 
     construct {
-        popover = new Gtk.Popover (popover_button);
+        popover = new Gtk.Popover (null);
 
         var label = new Gtk.Label (placeholder);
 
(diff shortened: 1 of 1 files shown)


src/Widgets/EntryPopover/Generic.vala
```diff
@@ -27,31 +28,37 @@ public abstract class Tasks.Widgets.EntryPopover.Generic<T> : Gtk.EventBox {
...
+        }
+        popover_button_box.add (label);
 
-        popover_button = new Gtk.MenuButton () {
-            label = placeholder,
+        var popover_button = new Gtk.MenuButton () {
+            child = popover_button_box,
             popover = popover,
-            image = new Gtk.Image.from_icon_name (icon_name, Gtk.IconSize.BUTTON),
             always_show_image = icon_name != null
```
ryonakano, line 44: Do we still need to set `always_show_image` since now `image` property is not set?

src/Widgets/EntryPopover/Generic.vala
```diff
@@ -12,9 +12,10 @@ public abstract class Tasks.Widgets.EntryPopover.Generic<T> : Gtk.EventBox {
     public string placeholder { get; construct; }
     public T value { get; set; }
 
-    private Gtk.MenuButton popover_button;
     private T value_on_popover_show;
 
+    private Gtk.EventControllerMotion motion_controller;
```
ryonakano, line 17: Do we keep this class being based on Gtk.EventBox until actual GTK4 migration?

(Both choices―keeping that and rebasing on another class while we're here―are fine, just I thought this class no longer need to be based on Gtk.EventBox since it now uses EventControllerMotion)

src/Widgets/EntryPopover/Generic.vala
danirabbit: Oops thanks for catching that!

src/Widgets/EntryPopover/Generic.vala
danirabbit: In my experience box won't capture certain events in GTK3 even with the motion controller, so I just left it as event box. It looks like it's working here though, so changed!

ryonakano (approved):
Looks good and works as expected!

PR #394, opened 2024-09-05
Add PopoverMenuitem

danirabbit:
Gtk.ModelButton is private in Gtk4


jeremypw (approved):
Code looks good ant works as expected

PR #395, opened 2024-09-14
Utils: Use GLib.TimeZone.identifier instead of deprecated Glib.TimeZone

jeremypw:


leonardo-lemos (approved):

tintou:
This is a good opportunity to specify the GLib requirements (`dependency('glib-2.0', version: '>=2.68')`)

tintou (approved):

PR #402, opened 2025-02-26
CI: bump EDS and run arm workflows on arm runner

davidmhewitt:
Haven't tested if this works yet, but this matches the EDS bump in Mail and Calendar and we don't ship the Flatpak version of this I guess.


io.elementary.tasks.json
```diff
@@ -222,4 +219,4 @@
             ]
         }
     ]
-}
+}
```
ryonakano, line 222: Tiny thing, but another missing EOF newline……

ryonakano:
Works confirmed by purging `io.elementary.tasks`, setting up Google Account using Evolution installed from apt, and then building & installing this branch with flatpak-builder.

Screenshot_elementary-8_2025-02-27_00:14:39

PR #406, opened 2025-03-08
ScheduledTaskListBox: GTK4 prep

danirabbit:
* Can't `get_children` so iterate by index instead
* `remove` instead of `destroy`
* Minor code style with namespaced constants
* use `child` property
* Can't get action group by name, so instead lookup action from the main window
* Can't set ellipsize on HeaderLabel since it's no longer a label subclass
* Remove unnecessary idles (review with hide whitespace changes)
* use dialog `present` and connect to `response` since we can't `run`


jeremypw (approved):
Looks good. Everything seems to work as normal.

